PR #1, opened 2019-05-10
Update README.md

BrenoDeMedeiros:
Adding a descriptive documentation in the README

Signed-off-by: BrenoDeMedeiros <breno@google.com>


README.md
```diff
@@ -1,2 +1,2 @@
 # authservice
-Prototype of external auth service
+An implementation of [Envoy](https://envoyproxy.io) [External Authorization](https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/ext_authz_filter), focused on delivering authN/Z for [Kubernetes](https://kubernetes.io) and [Istio](https://istio.io).
```
liminw: Do we want to mention Kubernetes? Istio supports many different platforms.

BrenoDeMedeiros: Well, some of the authentication mechanisms we intend to develop are specific to the Kubernetes control plane. That doesn't mean that the Istio-oriented solutions we develop are specific to Kubernetes.

What about if instead I said "authN/Z for Kubernetes and Istio" I said "authN/Z solutions for Cloud Native platforms, such as Istio and Kubernetes" ?

liminw: Sounds good!

BrenoDeMedeiros: I simplified it a bit to 'solutions for Istio and Kubernetes'. I think inverting the order is sufficient, while adding a reference to Cloud Native could cause confusion given that Istio also manages non-Cloud Native workloads; also some might associate Cloud Native to CNCF and Istio is not a project there. Let me know what you think.

liminw: That's a good point. I agree that mentioning CNCF here is causing more confusion. I am fine with the current description.

PR #7, opened 2019-09-25
Initial code providing limited authentication capabilities

nickrmc83:
Initial code providing limited authentication capabilities.

- Required credentials and configuration are compiled in at build time.
- Only a single IdP is supported.
- Code is **not** production ready and has not been **security hardened** and therefore should not be used outside of experimentation.


nickrmc83:
LGTM

PR #20, opened 2019-10-01
Add the prefix "Bearer " to the Authorization header

### LINKED ISSUE #6: authservice should use Bearer Authorization headers
When the authservice sets the token into the Authorization header for the app, it should follow the OAuth "Bearer Token Usage" standard https://tools.ietf.org/html/rfc6750#section-2.1. Specifically, the token value should be prepended with `Bearer ` (that's bearer with a capital B followed by a space character).

cfryanr:
Resolves #6


cfryanr:
@googlebot I signed it!

PR #22, opened 2019-10-01
New dockerfile for system-independent compilation

cfryanr:
There are link errors when compiled on MacOS, so make it easier
to compile in a system-independent way by providing a dockerfile
for building an image which contains all of the tools needed to
compile and test the project from an interactive shell inside
the container


cfryanr:
@googlebot I signed it!

PR #23, opened 2019-10-01
Fix the "docker" target in the Makefile to use Dockerfile.builder

cfryanr:
The "docker" target in the Makefile was using an outdated approach and referenced a `Dockerfile` that does not exist anymore


cfryanr:
@googlebot I signed it!

PR #25, opened 2019-10-02
The check()'s request has no scheme, so don't rely on it

### LINKED ISSUE #24: GRPC Check()'s request object has no scheme for request which includes authcode, causes authservice to go into redirect loop
When the OIDC server redirects the user back to the authservice with the authcode, envoy does not include the `scheme` in the `request` object passed to the `check()` grpc invocation.

You can see in the envoy code in [check_request_utils.cc](https://github.com/envoyproxy/envoy/blob/master/source/extensions/filters/common/ext_authz/check_request_utils.cc#L103) that it intends to read the scheme from the headers and include it in the request object.

However, you can also see in the logs of the sidecar envoy proxy that the header has no `scheme`, so the scheme ends up setting set to empty string in the request object.

```
[2019-10-02 17:07:39.415][21][trace][http] [external/envoy/source/common/http/http1/codec_impl.cc:514] [C1012] message complete
[2019-10-02 17:07:39.415][21][debug][http] [external/envoy/source/common/http/conn_manager_impl.cc:620] [C1012][S3038797389355712708] request headers complete (end_stream=true):
':authority', '35.184.251.112'
':path', '/oauth/callback?code=<REDACTED>&state=somestate'
':method', 'GET'
'cache-control', 'max-age=0'
'upgrade-insecure-requests', '1'
'user-agent', 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/77.0.3865.90 Safari/537.36'
'sec-fetch-mode', 'navigate'
'sec-fetch-user', '?1'
'accept', 'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3'
'sec-fetch-site', 'cross-site'
'referer', 'https://demo.login.run.pivotal.io/login'
'accept-encoding', 'gzip, deflate, br'
'accept-language', 'en-US,en;q=0.9'
'cookie', '__Host-acme-state-cookie=U_H1gEHrR2Q6ql4HUpnwg7b0FS-d-rBj4sr9FN8SPLLEWO5KoTsZDJA2encWbrI9nhlWv92RKAwNDS_Wox4dzNSxunds1PVKXTNGejhK1w'
'x-forwarded-for', '10.128.15.194'
'x-forwarded-proto', 'https'
'x-envoy-internal', 'true'
'x-request-id', 'eb56154d-3d36-9296-ae58-ba699589bb07'
'x-envoy-decorator-operation', 'authcode-sample-app.default.svc.cluster.local:8080/oauth*'
'x-istio-attributes', 'CksKGGRlc3RpbmF0aW9uLnNlcnZpY2UuaG9zdBIvEi1hdXRoY29kZS1zYW1wbGUtYXBwLmRlZmF1bHQuc3ZjLmNsdXN0ZXIubG9jYWwKSQoXZGVzdGluYXRpb24uc2VydmljZS51aWQSLhIsaXN0aW86Ly9kZWZhdWx0L3NlcnZpY2VzL2F1dGhjb2RlLXNhbXBsZS1hcHAKMQoYZGVzdGluYXRpb24uc2VydmljZS5uYW1lEhUSE2F1dGhjb2RlLXNhbXBsZS1hcHAKKgodZGVzdGluYXRpb24uc2VydmljZS5uYW1lc3BhY2USCRIHZGVmYXVsdApPCgpzb3VyY2UudWlkEkESP2t1YmVybmV0ZXM6Ly9pc3Rpby1pbmdyZXNzZ2F0ZXdheS03NTc1ZmY2NDc4LXJuc3NqLmlzdGlvLXN5c3RlbQ=='
'x-b3-traceid', 'e227b3e674b2fa1470ba2e837fddd97c'
'x-b3-spanid', '70ba2e837fddd97c'
'x-b3-sampled', '1'
'content-length', '0'
```

It is unclear why the request has no scheme here. We can see in our Chrome browser's debug tools that the redirect from the OIDC server was in fact an https URL, and that the browser followed that redirect by making a GET request to the https URL.

The [authservice code puts out a trace log](https://github.com/istio-ecosystem/authservice/blob/master/src/filters/oidc/oidc_filter.cc#L188-L191) which shows that the request's scheme is empty:

```
[2019-10-02 17:07:40.408] [console] [trace] Check
[2019-10-02 17:07:40.409] [console] [trace] Process
[2019-10-02 17:07:40.412] [console] [debug] Call from @10.16.0.10 to @10.16.0.18
[2019-10-02 17:07:40.412] [console] [trace] Process: checking handler for ://35.184.251.112-/oauth/callback?code=<REDACTED>&state=somestate
```

The result is that the [`if` statement's expression](https://github.com/istio-ecosystem/authservice/blob/583e8f7a5f7e0a419454e17c35cf6d09c51732e3/src/filters/oidc/oidc_filter.cc#L196-L201) always resolves to false, even in the case where we want it to resolve to true, causing the authservice to never trigger its code which is supposed to exchange the authcode for the token. Instead it starts the token flow from scratch, resulting in a redirect loop which gets halted by the browser.

cfryanr:
Unclear why the request has no scheme, so perhaps we can revisit this later. For now, skipping the scheme comparison seems okay since the app should be configured to only allow https traffic for the alpha release anyway.

Resolves #24

### DIFF
--- a/src/filters/oidc/oidc_filter.cc
+++ b/src/filters/oidc/oidc_filter.cc
@@ -194,9 +194,7 @@ google::rpc::Code OidcFilter::Process(
   auto path_parts = common::http::http::DecodePath(
       request->attributes().request().http().path());
   if (request->attributes().request().http().host() == callback_host &&
-      path_parts[0] == idp_config_.CallbackPath().path &&
-      request->attributes().request().http().scheme() ==
-          idp_config_.CallbackPath().scheme) {
+      path_parts[0] == idp_config_.CallbackPath().path) {
     return RetrieveToken(request, response, path_parts[1]);
   }
   return RedirectToIdP(response);
--- a/test/filters/oidc/oidc_filter_test.cc
+++ b/test/filters/oidc/oidc_filter_test.cc
@@ -230,7 +230,7 @@ TEST(OidcFilterTest, RetrieveToken) {
   ::envoy::service::auth::v2::CheckResponse response;
   auto httpRequest =
       request.mutable_attributes()->mutable_request()->mutable_http();
-  httpRequest->set_scheme("https");
+  httpRequest->set_scheme(""); // Seems like it should be "https", but in practice is empty
   httpRequest->set_host(callback_path.hostname);
   httpRequest->mutable_headers()->insert(
       {common::http::headers::Cookie, "__Host-acme-state-cookie=valid"});


cfryanr:
Resolves #24

PR #29, opened 2019-10-12
Additional config

### LINKED ISSUE #28: Allow configuration of more settings without recompiling
It should be possible for community members to try the upcoming alpha release without compiling the code or rebuilding the container image.

Therefore, let's allow the configuration (without needing to recompile or build a new container image) of the following:

- jwks
  - Why: The authservice currently does not work correctly unless this is configured, so let's allow people to configure it for the alpha release. This is the most important item in this issue for the alpha release.
- cryptor_secret
  - Why: When placing the authservice in the same deployment as an app and its envoy sidecar, and then scaling that pod's number of replicas, incoming requests will be load balanced between multiple replicas of the authservice, so all of those authservices need to be configured to use the same cryptor_secret so they can all read and write the same cookies.
- cookie_name_ and state_cookie_name_
  - Why: When using the authservice in the same deployment as an app, and using it for multiple deployments (i.e. multiple apps) behind the same gateway, the authservice of the first app and the authservice of the second app need to avoid stepping on each other's cookies.
- log level
  - Why: when community members who want to try to alpha are debugging their deployment, they will need to be able to turn up the log level to trace, and otherwise may wish to turn it down. Since it already defaults to trace, this maybe isn't super important. Allowing the user to turn the log level down would allow them to prevent logging potentially sensitive information, so it would be nice to have.
  - Note: The authservice should probably unconditionally complain in the log if it can't find its json config file, since there's a bit of a chicken and egg consideration in deciding what log level to use to log the fact that it failed to learn about what log level it should be be using.
- address
  - Why: While 127.0.0.1 is a nice default for the address on which the authservice should listen, it prevents users from deploying the authservice as a separate deployment, in which case they would need it to listen on 0.0.0.0. Since we plan to only recommend using the authservice in the same deployment (i.e. same pod) as your app for the alpha release, this isn't super important. Nice to have.
- port
  - Why: Just in case it conflicts with the port that their app is using. Pretty unlikely, so not super important for the alpha. Nice to have.

cfryanr:
Allow more things to be configured via the json file.

Resolves #28


peterhaochen47:
Closing this PR due to DCO checking not passing. Submitted another PR that passes the DCO check.

Issue #41, opened 2019-10-15
Put `__Host` back into the cookie names


cfryanr:
We accidentally removed the `__Host` prefix from the cookie names.  We should probably put that prefix back.  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#Cookie_prefixes for explanation of that prefix.

nickrmc83:
@cfryanr I'll fix in #13

nickrmc83:
Fixed as part of #42

PR #42, opened 2019-10-18
13 select token

### LINKED ISSUE #13: authservice should share both the ID token and the access token with the app
Currently the authservice fetches both the id token and the access token from the OIDC provider. It then saves the ID token into a browser cookie, and sets the ID token into the `Authorization` header for the app.

We believe that some apps would prefer to receive the access token, some apps may want to the ID token, and some apps may want both.

Our suggestion is to make this configurable.  One possible way to make this configurable would be to expose two new configuration options:

1. The name of the http header which the authservice should use to transmit the *access token* to the app. This could default to `Authorization`, since the `Authorization` header is typically used for access tokens for APIs. The content of this header would be `Bearer <token>` where `<token>` is the JWT token value, since this is the standard format for oauth tokens passed via the `Authorization` header.

2. The name of the http header which the authservice should use to transmit the *ID token* to the app. This could default to `X-AuthService-ID-Token`. Perhaps just to be consistent, this should also be passed in the `Bearer <token>` format.  That way, if the user overrides the default and configures this value to instead be `Authentication` then it would be correctly passed in the bearer token format.

If the user accidentally configures both of these options to be the same header name, then perhaps a reasonable behavior would be for the authservice to log a warning and then send only the ID token using that header name (or maybe log an error and abort).

nickrmc83:
- Added ability to optionally store and forward access tokens as well as identity tokens.
- Refactored cookie naming so that they always start with `__Secure-`
- Added configuration options for setting headers that will include id or access tokens

fixes #13


nickrmc83:
@cfryanr @peterhaochen47 @tylerschultz see changes here. They can't be merged until the Envoy multi-cookie fixes but apart from that this should be good to go pending review.

src/filters/oidc/oidc_filter.cc
```diff
@@ -182,26 +191,46 @@ google::rpc::Code OidcFilter::Process(
   }
    */
 
-  // Check if an authorization header already exists. If so let request
+  // Check if an id_token header already exists. If so let request
   // progress. It is up to the downstream system to validate the header is
   // valid.
   auto headers = request->attributes().request().http().headers();
-  if (headers.contains(common::http::headers::Authorization)) {
+  if (headers.contains(idp_config_.id_token().header())) {
```
cfryanr: @nickrmc83 I think what you're doing here is allowing the end-user to add the right header if they, for example, manually construct a curl to the app which the authservice is protecting.  That seems to make sense as a feature.  Now that the code is handling both ID Tokens and Access Tokens, would it make sense to skip token acquisition when the end user manually adds *either* the configured ID Token header *or* the configured Access Token header?

test/fixtures/valid-config.json
```diff
@@ -36,7 +36,14 @@
         "scopes": ["scope"],
         "landing_page": "page",
         "cryptor_secret": "some-secret",
-        "cookie_name_prefix": "my-app"
+        "cookie_name_prefix": "my-app",
+        "id_token": {
+          "preamble": "Bearer",
+          "header": "authorization"
+        },
+        "access_token": {
```
cfryanr: @nickrmc83 Do we understand correctly that this PR should be safe to merge as long as the user never configures "access_token" in their json config file?  We think that bug with setting multiple cookies will always actually set whichever `set-cookie` header was added last to the list of headers in the `check` response.  Since the ID Token header is added last, and the ID Token cookie is the only required cookie in the error checking code, it seems like it would work.

For the docs, based on your current PR code, we're thinking that we would document that the `id_token.header` is required and the `id_token.preamble`, while not technically required, should probably always be set to `Bearer` in practice. Does that sound right?  If `id_token.header` is required, than perhaps we should make it an error when the user fails to supply it?  Or would it be preferred to make both optional and make them default to `authorization` and `Bearer` when they are not supplied?

peterhaochen47:
@googlebot I consent

src/filters/oidc/oidc_filter.cc
nickrmc83: This is where I was looking for more input about what the expected behaviour should be. I'm happy to to skip acquisition if either are supplied.

nickrmc83:
@cfryanr I've introduced https://github.com/envoyproxy/protoc-gen-validate to allow config fields to be decorated with validation rules. This means we can drop the manual verification code and also generate documentation on allowable values.

config/common/BUILD
```diff
@@ -1,11 +1,22 @@
-cc_proto_library(
+load("@com_envoyproxy_protoc_gen_validate//bazel:pgv_proto_library.bzl", "pgv_cc_proto_library")
+#cc_proto_library(
```
nickrmc83: Delete

config/oidc/BUILD
```diff
@@ -1,12 +1,29 @@
...
+    cc_deps = ["//config/common:config_cc"],
+    linkstatic = True,
     visibility = ["//visibility:public"],
-    deps = ["//config/oidc:config_proto"],
+    deps = [
+        "//config/oidc:config_proto",
+    ],
 )
 
+#cc_proto_library(
```
nickrmc83: Delete

config/oidc/config.proto
```diff
@@ -3,17 +3,30 @@ syntax = "proto3";
...
+    string client_secret = 7 [(validate.rules).string.min_len = 1];
     repeated string scopes = 8;
-    string landing_page = 9;
-    string cryptor_secret = 10;
+    string landing_page = 9 [(validate.rules).string.min_len = 1]; // TODO: use [(validate.rules).string.uri_ref = true] when implemented
+    string cryptor_secret = 10 [(validate.rules).string.min_len = 1];
     string cookie_name_prefix = 11;
-}
+    TokenConfig id_token = 12 [(validate.rules).message.required = true];
+    TokenConfig access_token = 13;
```
nickrmc83: @cfryanr note `access_token` does not have a validation rule like `id_token`

src/filters/oidc/token_response.cc
```diff
@@ -35,17 +47,15 @@ absl::optional<TokenResponse> TokenResponseParserImpl::Parse(
...
-  auto token_type = fields.find("token_type");
+  auto token_type = fields.find(token_type_field);
   if (token_type == fields.end() ||
-      !(token_type->second.string_value() == "Bearer" ||
-        token_type->second.string_value() ==
-            "bearer")) {  // We allow either Bearer or bearer
+      !(token_type->second.string_value() == bearer_token_type)) {
     spdlog::info("{}: missing or incorrect `token_type` in token response",
                  __func__);
     return absl::nullopt;
```
peterhaochen47: @nickrmc83 Our test OIDC provider fails to pass this validation because it returns the lowercase `bearer` as `token_type`. We propose that we treat this `token_type` value in a case-insensitive manner (aka, both `Bearer` and `bearer` are okay).

We researched both OIDC and Oauth 2 specs, which do not conclusively dictate whether the `token_type` value is case sensitive or insensitive. Here are a few sources:

1. [Oauth 2 spec](https://tools.ietf.org/html/rfc6749#section-4.2.2):
```
   token_type
         REQUIRED.  The type of the token issued as described in
         Section 7.1.  Value is case insensitive.
```
Which says the value is case insensitive, which is also supported by the example given in [this source](https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/).

2. Meanwhile, the [OIDC spec](https://openid.net/specs/openid-connect-core-1_0.html#TokenResponse) says `The OAuth 2.0 token_type response parameter value MUST be Bearer, as specified in OAuth 2.0 Bearer Token Usage [RFC6750]`, which does indicates that it must be `Bearer`. But then again, the spec here is citing OAuth 2 spec, which, as mentioned above, says the value can be insensitive.

config/common/BUILD
nickrmc83: Removed in f7bc68875f3384b172dcecd24072987e42f22fb3

config/oidc/BUILD
nickrmc83: Removed in f7bc68875f3384b172dcecd24072987e42f22fb3

src/filters/oidc/token_response.cc
nickrmc83: I've reinstated the case insensitive comparison in f7bc68875f3384b172dcecd24072987e42f22fb3

test/fixtures/valid-config.json
nickrmc83: c2c026409ed98feceda56a8ede73b92a7e9e600b includes validation and specification of required fields leveraging protobuf annotations.

nickrmc83:
@cfryanr can you add a `lgtm` comment if you're now happy with all the changes in this PR?

fixes #41

peterhaochen47:
Also I think @nickrmc83 (the PR submitter) needs to approve our commit in order to make CLA pass: https://github.com/istio-ecosystem/authservice/pull/42#issuecomment-543996134

nickrmc83:
@googlebot I consent

cfryanr:
@googlebot I consent

Changdrew:
@googlebot I consent

PR #53, opened 2019-11-01
Issue35 async requests

sburton84:
These changes implement asynchronous request processing using multiple worker threads and not blocking on any requests to the upstream IDP.


bazel/BUILD.boost
```diff
@@ -0,0 +1,275 @@
...
+            "boost/context/**/*.hpp",
+        ],
+    ),
+    visibility = ["//visibility:public"],
+    deps = [
+        ":all",
+    ],
+)
+
+genrule(
```
nickrmc83: It looks like there's a fair amount of duplication in the various `gen_rules` for boost. Can we create a wrapper like https://github.com/istio-ecosystem/authservice/blob/master/bazel/bazel.bzl?

build/Dockerfile.builder
```diff
@@ -3,19 +3,30 @@ FROM debian:buster as bazel-builder
 COPY build/install-bazel.sh /build/
 RUN chmod +x /build/install-bazel.sh && /build/install-bazel.sh
 
-# Build auth binary.
+# Copy in only the necessary files for building.
 FROM bazel-builder as auth-builder
-COPY . /src
+COPY Makefile WORKSPACE /src/
```
nickrmc83: Can we collapse these `COPY` statements into 1 to reduce the number of layers.

build/Dockerfile.builder
```diff
@@ -3,19 +3,30 @@ FROM debian:buster as bazel-builder
...
 WORKDIR /src
 RUN bazel build //...
 
 # Create our final auth-server container image.
 FROM debian:buster
 RUN groupadd -r auth-server-grp && useradd -m -g auth-server-grp auth-server-usr
 
-COPY --from=auth-builder /src/bazel-bin/src/main/auth-server /app/auth-server
-RUN chgrp auth-server-grp /app/auth-server && chown auth-server-usr /app/auth-server && chmod u+x /app/auth-server
+COPY --from=auth-builder /src/bazel-bin/src/main/auth_server /app/auth_server
```
nickrmc83: Can we collapse these `COPY` statements into 1 to reduce the number of layers.

src/common/http/http.cc
```diff
@@ -349,6 +350,71 @@ response_t http_impl::Post(
...
+        const std::map<absl::string_view, absl::string_view> &headers,
+        absl::string_view body,
+        boost::asio::io_context& ioc,
+        boost::asio::yield_context yield) const {
+  spdlog::trace("{}", __func__);
+  try {
+    int version = 11;
+
+    ssl::context ctx(ssl::context::tlsv12_client);
+    // TODO: verify_peer should be used but is not currently working.
```
nickrmc83: Is this still true?

src/common/http/http.h
```diff
@@ -128,8 +131,21 @@ class http {
   virtual response_t Post(
       const authservice::config::common::Endpoint &endpoint,
       const std::map<absl::string_view, absl::string_view> &headers,
-      absl::string_view body)
-      const = 0;  // TODO: use string_view instead of const char *
+      absl::string_view body) const = 0;
+
+  /** @brief ASynchronously send a Post http message. To be used inside a Boost co-routine.
```
nickrmc83: s/S/s/

build/Dockerfile.builder
sburton84: I originally put the boost libraries into a single copy with a wildcard and it didn't seem to like it for some reason. I thought maybe it didn't like wildcard in `COPY --from` statements but it looks like I've used a wildcard in one in another project so I'll have another go and see if I can figure out why it's not working. Or at least I could just put all the separate filenames into a single copy.

build/Dockerfile.builder
sburton84: I originally tried this in one statement like this:
`COPY Makefile WORKSPACE bazel config src test /src/`
but the problem is that if any of the things on the left are directories instead of copying the directory it copies the *contents* of the directory, so you end up with all the contents of your subdirectories directly in the root `src` directory. Do you know any good way to do this in one statement without this happening?

src/common/http/http.cc
sburton84: I don't know, I didn't really touch any of the logic in here, what was the problem you were originally having with it?

nickrmc83: Previously using `verify_peer` instead of `verify_none` would result in an error which I cannot remember. We need to make sure it works now though else we have a massive security issue.

build/Dockerfile.builder
sburton84: Resolved this by using `.dockerignore` file to prevent the unnecessary files from being copied into the image.

src/common/http/http.cc
sburton84: Will try this out and see if I can fix it in a separate pull request after this one has been merged.

build/Dockerfile.builder
sburton84: Couldn't get it to work with a wildcard for some reason, it just omitted copying the files, but I've combined them all into a single COPY statement just listing each file separately.

src/service/async_service_impl.cc
```diff
@@ -0,0 +1,168 @@
...
+                  grpc::ServerCompletionQueue &cq, boost::asio::io_context &io_context)
+          : service_(service), cq_(cq), responder_(&ctx_), io_context_(io_context), config_(std::move(config)),
+            impl_(*config_) {
+    spdlog::trace("Creating processor state");
+    service.RequestCheck(&ctx_, &request_, &responder_, &cq_, &cq_, this);
+  }
+
+  void Proceed() override {
+    // Spawn a new instance to serve new clients while we process this one
+    new ProcessingState(config_, service_, cq_, io_context_);
```
cfryanr: Shouldn't this class hold a pointer to this object from the heap and delete it in its destructor? Maybe doesn't matter too much in practice because we know in practice that this is called by `main`. Still technically a memory leak though.  :)

src/service/async_service_impl.cc
```diff
@@ -0,0 +1,168 @@
...
+      if (!cq_->Next(&tag, &ok)) {
+        break;
+      }
+      GPR_ASSERT(ok);
+
+      static_cast<ServiceState *>(tag)->Proceed();
+    }
+  } catch (const std::exception &e) {
+    spdlog::error("{}: Unexpected error: {}", __func__, e.what());
+  }
```
cfryanr: Perhaps this could have a little more comments?  We're curious what conditions this is intended to handle.

It seems like you could arrive here if the previous loop caught any kind of exception or `Next` returned false.
  - If `Next` returned false in the first loop, would we expect it too also return false in the second loop?
  - On the other hand, if we arrive in the second loop because of an exception, should all kinds of exceptions lead to outstanding events being processed?

src/service/async_service_impl.cc
sburton84: This creates a new instance to handle any next request after the current one. We can't delete the state of the next request in the destructor of the current request. It's not really a leak because this handler for the next request is placed on the completion queue and will be later pulled off by `cq_->Next()` then processed and ultimately deleted by the destructor of `CompleteState`. This is the same way the [async gRPC example](https://github.com/grpc/grpc/blob/v1.24.0/examples/cpp/helloworld/greeter_async_server.cc) works.

Personally I don't really like using raw `new` like this either, but the way that async gRPC and its completion queue work makes it very difficult to avoid.

I'll try to add a few more comments to explain exactly what's happening here.

src/service/async_service_impl.cc
sburton84: This is mainly to work with a peculiarity of the gRPC completion queue. The second loop is just to drain the queue after shutting it down, because if the completion queue's destructor is called before draining any outstanding events it triggers an assertion. I'll add some additional comments to explain this.

nickrmc83:
/lgtm

Once the CI has passed I'll merge this PR.

Issue #54, opened 2019-11-04
Allow specification of TLS root certificates for an IdP


nickrmc83:
Currently any internal IdP whose TLS certificate is not signed by a well-known authority or some delegation thereof will stop working once we start verifying peers correctly. We should make sure it is possible to specify an optional CA certificate for verifying IdP endpoints.

stefanhenseler:
@cfryanr I'm wondering what the value must be for this option? Do I actually have to put the pem cert in there? I've tried the cacert fingerprint and the issuer name, but with no success.

cfryanr:
The pem formatted CA cert. Here's an (abbreviated) example from a configmap. Note that `\n` is newlines in json strings.

"trusted_certificate_authority": "-----BEGIN CERTIFICATE-----\nMIIE4jCCAsqgAwIBAgIBATANBgkqhkiG9w0BAQsFADARMQ8wDQYDVQQDEwZzZm8t\nY2EwHhcNMTkxMjIwMDAxNjI1WhcNMjEwNjIwMDAxNjIxWjARMQ8wDQYDVQQDEwZz\n...\nwPAYZhormzV5LxXMSd3BMdyexNvNiGffULhnYEebFI9GouFxV1LPdVu058LRW/db\n6PDm7+GEq/CcQhTgYOELmmcnC89zNxcCXiahxqKIMTuid295N4NldyK/IT4Tn4GN\nVknTT/Hr\n-----END CERTIFICATE-----"

stefanhenseler:
Thanks! that worked.

PR #55, opened 2019-11-07
Enable TLS peer verification and add CA root certificates to Docker images

sburton84:
Enables TLS peer verification for connection to upstream IDP services, to ensure the connections are secure. Also adds standard ca_certificates package to the Docker images which is required for verification of the certificate chains.


nickrmc83:
@cfryanr do you want to wait for #54 before merging this?

cfryanr:
> @cfryanr do you want to wait for #54 before merging this?

We’ve been testing against an IDP which has TLS certs signed by a trusted CA, so I don’t think it will effect us to merge this before #54 is implemented.

Issue #16, opened 2019-09-25
Support using refresh tokens to automatically refresh the access token before/when it expires


cfryanr:
The goal of this feature is to avoid giving an expired access token to the app for as long as possible by using the refresh token to fetch a new access token, until the refresh token itself expires, at which point the authservice would start a brand new token request.

One possible implementation could be as follows. If the authservice receives a refresh token along with the ID and access tokens, then it would:

- Save the refresh token into the encrypted browser cookie along with the other tokens
- There would be no need to send the refresh token to the app via a header, because the authservice is going to take care of using the refresh token on behalf of the app
- On each subsequent request, the authservice would decrypt the access token, ID token, and refresh token from the user's cookie and check them all to see which have expired
  - If none have expired, then proceed as usual
  - If either the ID token or the access token has expired, and the refresh token has also expired, then clear the cookie, start a brand new token request flow, and proceed as usual
  - If either the ID token or the access token has expired, but the refresh token is still good, then make a request to use the refresh token to get new ID and access tokens (and potentially a new refresh token too). Save the new tokens into the encrypted cookie, and set the access and ID tokens into the headers to pass them to the app.

cfryanr:
Closed by https://github.com/istio-ecosystem/authservice/pull/67

PR #66, opened 2020-02-05
169658500 refresh token

srwaggon:
Addresses issue #16 , refresh token.


srwaggon:
We are replacing this pull request with another that has authors/committers/sign-offers fixed.

PR #68, opened 2020-02-18
Instead of configuring a landing page, after auth automatically send the browser to the original page that it had requested

### LINKED ISSUE #43: Instead of configuring a `landing_page`, could the authservice send the user back to the original request path that initiated the oauth authorize flow?
If I visit `example.com/some_path` and the authservice decides that it needs to redirect me to an IDP to login and get tokens, could the authservice send me back to `example.com/some_path` after it obtains the tokens?

At first glance it would appear that this would require some kind of state to be stored in the authservice. But is there another way?  For example, could this state be put into another browser cookie to avoid storing the state server-side?

cfryanr:
Fixes #43


cfryanr:
@googlebot I fixed it.

Issue #91, opened 2020-04-09
Authservice should not drop requests when invoking callback url after session expiration


tylerschultz:
If a user were to take too long to log in (a time period longer than the session timeout), the user's request will be dropped by envoy leaving the user waiting for the authservice to redirect to the app url, which will never happen. The auth service should return an http response explaining what has happened.

PR #92, opened 2020-04-09
Envoy to return 400 status instead of dropping request

tylerschultz:
Cause Envoy to return 400 status code when accessing callback url
when session is expired. This may happen if the user is redirected to
the IdP and takes more time to log in than the session timeout

Fixes https://github.com/istio-ecosystem/authservice/issues/91

Signed-off-by: Tyler Schultz <tschultz@pivotal.io>


margocrawf:
@googlebot I fixed it.

margocrawf:
@googlebot I consent.

PR #100, opened 2020-06-09
Upgrade ssl context to sslv3

AndrewBabbitt97:
tlsv12 causes issues with validating certificates on some idP servers with valid ssl certs, upgrading to sslv3 fixes these issues.

### DIFF
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -363,7 +363,7 @@ response_t HttpImpl::Post(absl::string_view uri,
   try {
     int version = 11;
 
-    ssl::context ctx(ssl::context::tlsv12_client);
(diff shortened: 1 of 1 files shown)


AndrewBabbitt97:
@googlebot I signed it!

AndrewBabbitt97:
@googlebot I signed it!

AndrewBabbitt97:
@googlebot I signed it!

AndrewBabbitt97:
Disregard the change, however of note is that it looks like the version of debian your last release was built with has an issue with some ssl certs, found the reason my change worked wasnt the code change but by building the image using a newer version of the debian image.

PR #105, opened 2020-06-23
Enhance boost::asio strategy 

Changdrew:
Allow for every asio call to resume on any available worker thread


Changdrew:
@gogolebot I consent.

Changdrew:
@googlebot I consent.

margocrawf:
@googlebot I consent.

Changdrew:
@cfryanr Could you take a look at this if/when you have time? We'd love your input. 

We are unsure of how to obtain test coverage of filter chain matching with respect to the refactor.

cfryanr:
Hi @Changdrew and @margocrawf,

Nice job.

While looking at the diffs to understand what you had done, I noticed some dead code, so I pushed a commit to remove the dead code.

It looks like you're asking about the lost test coverage from when you removed service_impl_test.cc. It appears that you could bring it back as async_service_impl_test.cc and adjust those tests to make them unit tests of your new `Check` function. In order to call your new `Check` function, you need the following arguments to be created by your test:

- a request and a response, which the old test had
- a vector of `FilterChain`: it seems like you could create an instance of `config::FilterChain` by calling its default constructor and then calling its setters to fill in the minimum required stuff, then you could pass it to the constructor of `FilterChainImpl` to make one of those, and then stick that into a vector.
- `config::TriggerRule`: seems like you could make one with the default constructor and then call the setters to add the same trigger rules that are in `valid-config.json`.
- an `ioc` and a `yield`: [check out this example from my new commit](https://github.com/istio-ecosystem/authservice/blob/0d3db2dffbdd331276be811e15febec0b0aba78c/test/filters/oidc/oidc_filter_test.cc#L189) of making `oic` and `yield` objects for another test by writing a simple test helper function

Once you've got all those things created in your test setup, then you can call `Check`. It appears that those tests just wanted to vary the request url and confirm that the configured trigger rules caused the right responses.

Let me know if that helps, or if you need any help trying that.

Thanks!

cfryanr:
@googlebot I consent!

PR #119, opened 2020-09-17
Update Istio download script to use 1.4.10

cflewis:
1.4.0 was causing Envoy crashes following the bookinfo example. Using 1.4.10 prevented the Envoy crashes.


cflewis:
P.S. I strongly suggest testing this on your side before merging. It is entirely possible I missed some sort of step when following the doc, but this change appears innocuous given that the minor version remains the same.

cflewis:
I have changed my mind on this, I actually haven't been able to successfully deploy the example, so I can't be sure it works at all. Sorry for the noise.

PR #120, opened 2020-09-21
Fix Run: Unexpected error: !ok

### LINKED ISSUE #112: Run: Unexpected error: !ok
So we have been experiencing issues with gRPC shutting down in authservice as people authenticate, would result in 403s from envoy as it could no longer communicate to authservice while the actual process continued to run only way to resolve would be to manually delete/recycle the authservice pod. I have added logs with trace logging for authservice (with gRPC tracing) and envoy to the issue.

Commenting out the break on line 137 of async_service_impl.cc fixes the issue (the logs uploaded are using this fix). It looks like we were exiting on a bool that isnt fatal if its false thus causing the server to shut down.
Fix

The authservice log is the interesting one with error entries for gRPC as follows:

"Attempt to send initial metadata after stream was closed"
"Attempt to send message after stream was closed"
"Attempt to send trailing metadata after stream was closed"
 
However in terms of the browser request, it doesnt even appear to have any issues and people are able to auth without issues with this change.

Looking at the gRPC documentation it looks as if Next() returns a bool that if false you should shut the server down, but doesnt seem to be the case for the ok value.

Bear in mind haven't personally worked with gRPC on any projects and could be totally wrong on this fix, however if this is an acceptable fix i can submit a PR for it.

I am however still curious as to what is going on with the stream closing errors, as I'm not seeing anything in the envoy logs relating to it.

[authservice.log](https://github.com/istio-ecosystem/authservice/files/4901751/authservice.log)
[envoy.log](https://github.com/istio-ecosystem/authservice/files/4901752/envoy.log)

AndrewBabbitt97:
Fixes #112

Not going to go into details here, please reference original issue.

We have been running this change in prod for a while now with no issues.

Only question is if we want to just remove that block of code entirely or leave that logging as is.

### DIFF
--- a/src/service/async_service_impl.cc
+++ b/src/service/async_service_impl.cc
@@ -183,7 +183,6 @@ void AsyncAuthServiceImpl::Run() {
       // tells us whether there is any kind of event or cq_ is shutting down.
       if(!ok) {
         spdlog::error("{}: Unexpected error: !ok", __func__);
-        break;
       }
 
       static_cast<ServiceState *>(tag)->Proceed();


kfud:
When is this fix going to be release?

AndrewBabbitt97:
Not sure if it will at this rate, devs haven't been active on this project at all the past few months. I can see about reaching out to some other Istio devs to see how this projects going to be handled.

kfud:
I can't even make the project due to vague dependency errors, it's too bad because it's a great authentication proxy.

kfud:
I can confirm this fix works... auth-service has been running stable for a good week now, where as before it would crash every hour.

Issue #122, opened 2020-11-09
Istio 1.7.x and restarts istio-ingressgateway


bocha-fanatic:
**Description**

istioctl version
```
no running Istio pods in "istio-system"
1.7.4
```

istioctl operator init
```
Using operator Deployment image: docker.io/istio/operator:1.7.4
✔ Istio operator installed
✔ Installation complete
```

kubectl create ns istio-system

cat istio.yaml
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
metadata:
  name: istio
  namespace: istio-system
spec:
  profile: default
  values:
    global:
      proxy:
        clusterDomain: "test.domain.local"
#        logLevel: warning
#        componentLogLevel: "misc:error"
      trustDomain: "test.domain.local"
    gateways:
      istio-ingressgateway:
        type: LoadBalancer
        loadBalancerIP: 10.11.0.112
```

kubectl apply -f istio.yaml
`istiooperator.install.istio.io/istio created`

cat configmap-keycloak.yaml
```
---
kind: ConfigMap
apiVersion: v1
metadata:
  name: keycloak
  namespace: istio-system
data:
  config.json: |
    {
      "listen_address": "127.0.0.1",
      "listen_port": "10003",
      "log_level": "info",
      "threads": 8,
      "chains": [
        {
          "name": "idp_filter_chain",
          "filters": [
          {
            "oidc":
              {
                "authorization_uri": "https://{keycloak:port}/auth/realms/master/protocol/openid-connect/auth",
                "token_uri": "https://{keycloak:port}/auth/realms/master/protocol/openid-connect/token",
                "callback_uri": "https://{keycloak:port}/oauth/callback",
                "jwks": "{...}",
                "client_id": "{...}",
                "client_secret": "...",
                "trusted_certificate_authority": "{...}",
                "scopes": [],
                "cookie_name_prefix": "",
                "id_token": {
                  "preamble": "Bearer",
                  "header": "Authorization"
                },
                "access_token": {
                  "preamble": "Bearer",
                  "header": "Authorization"
                },
                "logout": {
                  "path": "/authservice_logout",
                  "redirect_uri": "https://{keycloak:port}/auth/realms/master/protocol/openid-connect/logout"
                }
              }
            }
          ]
        }
      ]
    }
```

kubectl apply -f configmap-keycloak.yaml
`configmap/keycloak created`

kubectl edit deployments -n istio-system istio-ingressgateway
[...]

PR #123, opened 2020-11-11
Fixes envoy filter to work with istio 1.7.4

stefanhenseler:
Fixes issue #122. Seems like the typed config makes the filter work again with istio 1.7.x. I've used the example here: https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/ext_authz_filter

### DIFF
--- a/bookinfo-example/config/productpage-external-authz-envoyfilter-sidecar.yaml
+++ b/bookinfo-example/config/productpage-external-authz-envoyfilter-sidecar.yaml
@@ -17,43 +17,45 @@ spec:
     labels:
       app: productpage # target the app to which you want to apply the filter
   configPatches:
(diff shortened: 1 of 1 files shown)


stefanhenseler:
@googlebot I signed it!

stefanhenseler:
> @googlebot I signed it!

ipuustin:
I think the `EnvoyFilter` should be gradually deprecated, because the CUSTOM action in `AuthorizationPolicy` seems to be a less fragile way of doing this: https://istio.io/latest/docs/tasks/security/authorization/authz-custom/

Have you tried it out with AuthService?

incfly:
This is done. Please check the book-info/readme.md now. thanks!

PR #126, opened 2021-01-06
Add option to skip TLS verification in OIDC filter

bruegth:
In case of an internal OIDC Provider behind envoy, the verification of the certificate may not needed.

### DIFF
--- a/config/oidc/config.proto
+++ b/config/oidc/config.proto
@@ -171,4 +171,9 @@ message OIDCConfig {
     // When specified, the Authservice will use the configured Redis server to store session data.
     // Optional.
     RedisConfig redis_session_store_config = 16;
+
+    // When specified, the Authservice will not validate OIDC Identity Provider certificate.
+    // Default is false.
+    // Optional.
+    bool skip_openid_provider_tls_verify = 17;
 }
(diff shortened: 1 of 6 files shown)


bruegth:
@googlebot I fixed it.

incfly:
OIDC is built on top of OAuth2 which mandates HTTPS AFAIK https://datatracker.ietf.org/doc/html/rfc6749#section-3.1

Prod environment can have insider compromise and we don't want the employee(or other user account) be transmitted over plain-text either. If public cert is not available to you, you can use your own organization PKI(self signed), and have authservice use plugged root for verification (which makes sense).

bruegth:
For my use case it makes no sense to put an certificate on the IDP (Keycloak) because I put all components within a single VM and they are all covered by an envoy proxy which terminates the TLS. 
Also many other OIDC handler supporting an option to skip TLS verification.
But I'm fine with your comment and I see I'm not on the RFC side -> so closing this PR...

incfly:
> Also many other OIDC handler supporting an option to skip TLS verification.

Purely curious, any doc/reference of these OIDC handlers?

bruegth:
> > Also many other OIDC handler supporting an option to skip TLS verification.
> 
> Purely curious, any doc/reference of these OIDC handlers?

e.g.
https://github.com/louketo/louketo-proxy/blob/master/config.go#L64

PR #127, opened 2021-01-06
Let pass websocket requests with access token in as query parameter too

bruegth:
See here:
https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api

Or here:
https://www.envoyproxy.io/docs/envoy/latest/configuration/http/http_filters/jwt_authn_filter.html?highlight=jwt#default-extract-location


bruegth:
No review after one year closing this PR. 
Seems no one except of me is using Websockets with this library... :(

PR #130, opened 2021-01-23
auth message V3 migration and backward compatibility

### LINKED ISSUE #128: Depreciated V2 update to External Authz Service V3?
V2 will be disabled in Q1/2021.

Update to V3 seems not so easy to me isn't it?

### LINKED ISSUE #129: Too-old bazel can be required to build
It requires too old Bazel (such as 1.0.0) to build executable. It is because this project hires `--incompatible_depset_is_not_iterable` flag which deprecated newer releases. This flag seems only migration solution and destroyed a long time ago. https://github.com/bazelbuild/bazel/commit/9f9f5ec10fed81362c35c50eeb4515bbe5bf8bce

Shikugawa:
close #128
close #129

This PR auth message API version to V3 with upgrading envoy v1.17.0. Following with newer version of the envoy, we fixed

1. Fix bazel version to use 3.7.2
2. lot of changes of build definition to build authservice successfully.

And, this PR introduces breaking change of message version. We should discuss to accept or not this breaking V2 deprecation.


Shikugawa:
@ZackButcher @lizan ping

.bazelrc
```diff
@@ -7,3 +7,6 @@
 # The following workaround was suggested by https://github.com/grpc/grpc/issues/19645
 #
 build --copt -DGRPC_BAZEL_BUILD
+build --cxxopt=-std=c++17
+build --test_output=all
```
lizan, line 11: I wouldn't add this here as it would be too noisy running locally, if you want this on CI add there.

bazel/repositories.bzl
```diff
@@ -0,0 +1,146 @@
...
+    strip_prefix = "boringssl-936ca21922d266a31e3309144b082bdb3a689af7",
+    urls = [
+      "https://github.com/google/boringssl/archive/936ca21922d266a31e3309144b082bdb3a689af7.tar.gz"
+    ],
+  )
+
+def zlib():
+  http_archive(
+    name = "zlib",
+    build_file = "//bazel:zlib.BUILD",
```
lizan, line 106: You should be able to use zlib BUILD file from protobuf so you don't need this file.

Shikugawa: `no such package '@zlib//': The repository '@zlib' could not be resolved and referenced by '//external:madler_zlib'` message had occurred on my environment.

lizan: That's because you don't have `protobuf_deps()` called in WORKSPACE. https://github.com/protocolbuffers/protobuf/blob/master/protobuf_deps.bzl

Shikugawa: `protobuf_deps()` can be called on `grpc_extra_deps()` https://github.com/grpc/grpc/blob/v1.34.0/bazel/grpc_extra_deps.bzl#L29
And this is already called in WORKSPACE. Does it means we should download protobuf explicitly and calls `protobuf_deps()`?

Shikugawa:
@lizan PTAL

lizan (approved):

Issue #132, opened 2021-01-26
Configuration secret isn't DRY and can exceed K8s annotation/secret size limits


atoy3731:
We utilize authservice behind a lot of different virtual services, however the way the current configuration file exists, we end up needing to create separate chains for every application, even though nearly all the contents are repeated. Here's a redacted version of the configuration file with 2 applications, containing the same values across both:

```
{
	"listen_address": "0.0.0.0",
	"listen_port": "10003",
	"log_level": "trace",
	"threads": 8,
	"chains": [
		{
			"name": "chain-1",
			"match": {
				"header": ":authority",
				"prefix": "foo.example.com"
			},
			"filters": [
				{
					"oidc": {
						"authorization_uri": "https://keycloak.example.com/auth/realms/example-realm/protocol/openid-connect/auth",
						"token_uri": "https://keycloak.example.com/auth/realms/example-realm/openid-connect/token",
						"callback_uri": "https://foo.example.com/login",
						"jwks": "MATCHES-ACROSS-CHAINS",
						"client_id": "client-id-1",
						"client_secret": "client-password-1",
						"scopes": [],
						"cookie_name_prefix": "foo-cookie-prefix",
						"id_token": {
							"preamble": "Bearer",
							"header": "Authorization"
						},
						"access_token": {
							"header": "JWT"
						},
						"logout": {
							"path": "/logout",
							"redirect_uri": "https://keycloak.example.com/auth/realms/example-realm/protocol/openid-connect/token/logout"
						}
					}
				}
			]
		},
		{
			"name": "chain-2",
			"match": {
				"header": ":authority",
				"prefix": "bar.example.com"
			},
			"filters": [
				{
					"oidc": {
						"authorization_uri": "https://keycloak.example.com/auth/realms/example-realm/protocol/openid-connect/auth",
						"token_uri": "https://keycloak.example.com/auth/realms/example-realm/openid-connect/token",
						"callback_uri": "https://bar.example.com/login",
						"jwks": "MATCHES-ACROSS-CHAINS",
						"client_id": "client-id-1",
						"client_secret": "client-password-1",
						"scopes": [],
						"cookie_name_prefix": "bar-cookie-prefix",
						"id_token": {
							"preamble": "Bearer",
							"header": "Authorization"
						},
						"access_token": {
							"header": "JWT"
						},
						"logout": {
							"path": "/logout",
[...]

PR #133, opened 2021-02-04
Add global filter chain configuration API part 1

Shikugawa:
The initial part of adding global filter chain configuration. #132

It allows for users to set global configuration that will be applied to all filters in the same filter chain.

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -53,6 +53,11 @@ message FilterChain {
     // Optional.
     Match match = 2;
 
+    // Global configuration of this filter chain. It will be dismissed when OIDC config set per filter.
+    // Configuration items will be shared with all filters.
+    // Optional.
+    oidc.OIDCConfig global_config = 4;
+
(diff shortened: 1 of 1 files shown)


config/config.proto
```diff
@@ -53,6 +53,11 @@ message FilterChain {
     // Optional.
     Match match = 2;
 
+    // Global configuration of this filter chain. It will be dismissed when OIDC config set per filter.
```
ZackButcher, line 56:
```suggestion
    // Global configuration of this filter chain. It will be ignored when OIDC config set per filter.
```

I assume it's either global or individual today, with no support for overrides yet?

ZackButcher (approved):

config/config.proto
Shikugawa: I think that it should be overridden by individual config if that exists.

ZackButcher (approved):

lizan (approved):

PR #136, opened 2021-02-19
Add global filter chain configuration API part 2

### LINKED ISSUE #132: Configuration secret isn't DRY and can exceed K8s annotation/secret size limits

Shikugawa:
Close #132


Shikugawa:
cc: @ZackButcher @lizan

ZackButcher:
Can you update https://github.com/istio-ecosystem/authservice/blob/master/docs/README.md (which looks generated, but I haven't looked into how yet), and add an example using the override somewhere so it's easy for someone to see how to use it?

test/filters/filter_chain_test.cc
```diff
@@ -87,5 +88,32 @@ TEST(FilterChainTest, New) {
...
 }
 
+TEST(FilterChainTest, Override) {
+  auto default_config = config::oidc::OIDCConfig();
+  default_config.set_authorization_uri("https://istio.io/auth/default");
+  default_config.set_token_uri("https://istio.io/token");
+  default_config.set_jwks("default_jwk");
+
+  auto configuration =
+      std::unique_ptr<config::FilterChain>(new config::FilterChain);
```
lizan, line 98: nit: make_unique

cfryanr:
> Can you update https://github.com/istio-ecosystem/authservice/blob/master/docs/README.md (which looks generated, but I haven't looked into how yet), and add an example using the override somewhere so it's easy for someone to see how to use it?

In case it helps, you can regenerate the docs with `make docs` and then commit the updated doc. See [Makefile](https://github.com/istio-ecosystem/authservice/blob/master/Makefile#L10-L13).

lizan (approved):

PR #137, opened 2021-03-16
Introduce Lazy OIDCConfig validation

Shikugawa:
In the current implementation, `default_oidc_config` and `oidc_override` will be validated before starting the server, it results in users can't configure them without stripping redundant information, they can't be DRY.

In this patch, we fix this problem by delaying config evaluation after all of fragmented config will be merged.


config/oidc/config.proto
```diff
@@ -172,3 +172,120 @@ message OIDCConfig {
     // Optional.
     RedisConfig redis_session_store_config = 16;
 }
+
+// This OIDC config doesn't need to be validate. It is introduced to use with `default_oidc_config` and `oidc_override`
+// LooseOIDCConfig will be merged to construct OIDC Config and validated.
+message LooseOIDCConfig {
```
lizan, line 178: Not fan of this kind of copy proto. You should be able to control where PGV kicks in in the code.

Shikugawa:
@lizan PTAL

src/config/get_config.cc
```diff
@@ -18,66 +18,60 @@ using namespace google::protobuf::util;
...
+}
+
+void ConfigValidator::ValidateUri(absl::string_view uri,
+                                  absl::string_view uri_type,
+                                  absl::string_view required_scheme) {
+  std::unique_ptr<common::http::Uri> parsed_uri;
   try {
-    parsed_uri = unique_ptr<common::http::Uri>(new common::http::Uri(uri));
-  } catch (runtime_error &e) {
+    parsed_uri = std::unique_ptr<common::http::Uri>(new common::http::Uri(uri));
```
lizan, line 52: nit: std::make_unique

lizan (approved):

PR #139, opened 2021-03-24
Add a new filter type: "mock"

ipuustin:
Add a "mock" filter type. The filter approves or denies requests based on configuration value. The benefit of this is twofold:

1. Allow AuthService to run without OIDC providers. This can be needed if the AuthService configuration is dynamically constructed and may sometimes be without any OIDC chains.
2. Allow testing of external authorization logic in the cluster without having to introduce and configure OIDC providers.

Also, this helps to keep the filter abstraction in working shape by preventing configuration validation etc. getting tied to OIDC.

### DIFF
--- a/src/filters/filter_chain.cc
+++ b/src/filters/filter_chain.cc
@@ -11,6 +11,7 @@
 #include "src/filters/oidc/oidc_filter.h"
 #include "src/filters/oidc/redis_session_store.h"
 #include "src/filters/pipe.h"
+#include "src/filters/simple/simple_filter.h"
 
 namespace authservice {
 namespace filters {
@@ -52,6 +53,9 @@ std::unique_ptr<Filter> FilterChainImpl::New() {
   for (auto &filter : *config_.mutable_filters()) {
     if (filter.has_oidc()) {
       ++oidc_filter_count;
+    } else if (filter.has_simple()) {
+      result->AddFilter(FilterPtr(new simple::SimpleFilter(filter.simple())));
+      continue;
     } else {
       throw std::runtime_error("unsupported filter type");
     }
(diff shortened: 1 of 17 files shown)


src/filters/filter_chain.cc
```diff
@@ -52,6 +53,9 @@ std::unique_ptr<Filter> FilterChainImpl::New() {
   for (auto &filter : *config_.mutable_filters()) {
     if (filter.has_oidc()) {
       ++oidc_filter_count;
+    } else if (filter.has_simple()) {
+      result->AddFilter(FilterPtr(new simple::SimpleFilter(filter.simple())));
```
Shikugawa, line 57: nit: `std::make_unique<SimpleFilter>(filter.simple())`

test/filters/filter_chain_test.cc
```diff
@@ -88,5 +88,16 @@ TEST(FilterChainTest, New) {
   ASSERT_TRUE(dynamic_cast<Pipe *>(instance.get()) != nullptr);
 }
 
+TEST(FilterChainTest, Simple) {
+  auto configuration =
```
Shikugawa, line 92: nit: `make_unique`

Shikugawa (changes requested):
Thanks! It generally LGTM but `simple` name is ambiguous a bit. Could you rename it? I think `mock` should be better.

ipuustin:
@Shikugawa thanks for the review! I implemented the changes you requested.

Shikugawa:
@ipuustin LGTM. Could you rewrite PR title into `Add a new filter type: "mock"`?

Shikugawa:
ping @lizan

lizan (approved):

PR #145, opened 2021-08-04
Adding working examples for istio-1.10.3

anhdle14:
This PR is not ready yet, but provide the first draft for updating the documentation to the latest version of Istio.

### DIFF
--- a/bookinfo-example/config/bookinfo-authn-policy-template-adding-reviews.yaml
+++ b/bookinfo-example/config/bookinfo-authn-policy-template-adding-reviews.yaml
@@ -2,21 +2,32 @@
 # A simple example of enabling authentication for some services.
 # See docs: https://istio.io/docs/tasks/security/authentication/authn-policy/#end-user-authentication
 #
-
 ---
-apiVersion: "authentication.istio.io/v1alpha1"
-kind: "Policy"
+apiVersion: security.istio.io/v1beta1
+kind: RequestAuthentication
 metadata:
-  name: "bookinfo-authn-policy"
+  name: "bookinfo-request-authn"
   namespace: default
 spec:
-  targets:
-    # Enable authentication for the following services.
-    # Note: `request.auth` only appears on the ServiceRoleBinding's or AuthorizationPolicy's for the following services
-    - name: productpage
-    - name: reviews
-  origins:
-    - jwt:
-        issuer: https://demo.example.replace.me.com/oauth/token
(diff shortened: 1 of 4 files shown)


ZackButcher (commented):
Looks great, but we can get rid of the `EnvoyFilter` entirely if we do... (following along from the [new external auth docs](https://istio.io/latest/docs/tasks/security/authorization/authz-custom/)).
1. Create a ServiceEntry so Istio knows about authservice (only needed if it's run on localhost; if it's run deployed as a service in k8s we don't need to write this ourselves):
    ```yaml
    apiVersion: networking.istio.io/v1alpha3
    kind: ServiceEntry
    metadata:
      name: external-authz-grpc-local
      namespace: authservice
    spec:
      hosts:
      - "external-authz-grpc.local" # The service name to be used in the extension provider in the mesh config.
      ports:
      - name: grpc
        number: 10003 # The port number to be used in the extension provider in the mesh config.
        protocol: GRPC
      resolution: STATIC
      endpoints:
      - address: "127.0.0.1"
    ```

1. [Let Istio know about authservice as an ext_authz provider](https://istio.io/latest/docs/tasks/security/authorization/authz-custom/#define-the-external-authorizer), something like:
    ```yaml
    data:
      mesh: |-
        # Add the following content to define the external authorizers.
        extensionProviders:
        - name: "ext-authz-local-grpc"
          envoyExtAuthzGrpc:
            service: "external-authz-grpc-local.authservice" # name of the ServiceEntry we created in 1
            port: "10003"
   ```

1. Update the AuthorizationPolicy above to use `action: CUSTOM` rather than `action: ALLOW`:
    ```yaml
    apiVersion: security.istio.io/v1beta1
    kind: AuthorizationPolicy
    metadata:
      name: "bookinfo-authz-policy"
    spec:
      selector:
        matchLabels:
          app: productpage
          app: reviews
      action: CUSTOM
      provider:
        name: ext-authz-local-grpc # name from `mesh` declaration in 2
      rules:
        - from:
            - source:
                requestPrincipals: ["*"]
    ```

anhdle14:
Thank you for the suggestion. I will make some new changes.

One thing I would like to clarify is the example for auth-service in gateway. I try to adopt the logic to the gateway as well but doesn't work. The site will only get `RBAC: access denied` message. If possible I would like to update README.md for that or write a proper example for it.

anhdle14:
Close in favor of #148

PR #147, opened 2021-08-13
Update authservice using external authz.

incfly:
This PR updates a few things.

- Use external authz provider, the CUSTOM action.
- Change to use credentialName instead of mounting secrets for ingress to host the OIDC callback URI.
- Other misc polishment such as logging the jwks parsing error and makefile IMAGE customization.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ SRCS=$(shell find . -name '*.cc')
 HDRS=$(shell find . -name '*.h')
 TARGET:=//src/main:auth_server
 BAZEL_FLAGS:= --verbose_failures
+IMAGE?=authservice:$(USER)
 
 all: build test docs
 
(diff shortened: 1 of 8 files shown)


lizan:
Can you fix DCO?

Makefile
```diff
@@ -18,7 +19,7 @@ compose:
 	docker-compose up --build
 
 docker: build
-	rm -rf build_release && mkdir -p build_release && cp -r bazel-bin/ build_release && docker build -f build/Dockerfile.runner -t authservice:$(USER) .
+	rm -rf build_release && mkdir -p build_release && cp -r bazel-bin/ build_release && docker build . -f build/Dockerfile.runner -t $(IMAGE) && docker push $(IMAGE)
```
Shikugawa, line 22: Should we push after build? I think that we may add `docker: push` stage to push them.

Shikugawa (commented):
LGTM with a nit comment. Thanks!

Makefile
incfly: remove. will add it later.

lizan (approved):

PR #148, opened 2021-08-14
Make authservice config as a Helm package.

incfly:
- Clarify the instructions on idp expectations,  reasons why we need to use localhost, etc.
- External Authz as a separate service.
- Helm package containing param for customization.
- clang-format fix.

https://github.com/istio-ecosystem/authservice/issues/149


bookinfo-example/auth.yaml
```diff
@@ -0,0 +1,211 @@
...
+                "client_id": "159575789034-27l81afqk927eg7urur2etntg91rv8on.apps.googleusercontent.com",
+                "client_secret": "5GHnk4Fs94x13Oif4hKhYV2d",
```
Shikugawa, line 26: I think these sections should be `<CLIENT_ID>` and `<CLIENT_SECRET>` because these are user-specific sections.

incfly: removed the file.

this should be accidentally added to the git. already part of the authservice helm packages.

lizan (approved):

PR #150, opened 2021-08-16
implement periodic JWKs fetcher/updater

Shikugawa:
Implement to fetch JWKs periodically and update them.


config/oidc/config.proto
```diff
@@ -75,14 +75,18 @@ message OIDCConfig {
...
+        // The JSON JWKS response from the OIDC provider’s `jwks_uri` URI which can be found in
+        // the OIDC provider's
+        // [configuration response](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
+        // Note that this JSON value must be escaped when embedded in a json configmap
+        // (see [example](https://github.com/istio-ecosystem/authservice/blob/master/bookinfo-example/config/authservice-configmap-template.yaml)).
+        // Used during token verification.
+        // Required.
+        string jwks = 4;
+        
+        string jwks_uri = 17;
```
incfly, line 88: comment doc string for this? semantics as well as how we handle it ("periodic refresh etc")

src/filters/oidc/jwks_storage.h
```diff
@@ -0,0 +1,75 @@
...
+  JwksStorage(const std::string& jwks, google::jwt_verify::Jwks::Type type) {
+    updateJwks(jwks, type);
+  }
+
+  const google::jwt_verify::JwksPtr& jwks() const {
+    std::unique_lock<std::mutex> lck(mux_);
+    return jwks_;
+  }
+
+  void updateJwks(const std::string& new_jwks,
```
incfly, line 30: will this only be invoked once instead of periodic refresh?

Shikugawa: No. In refactored implementation, `NonPermanentJwksStorage` wiil update existing JWKs but `PermanentJwksStorage` won't do it.

config/oidc/config.proto
Shikugawa: Yes. I will add detail docs in the future.

config/oidc/config.proto
```diff
@@ -75,14 +75,30 @@ message OIDCConfig {
...
+    // This message defines a setting to allow asynchronous retrieval and update of the JWK for
+    // JWT validation at regular intervals.
+    message JwksFetcherConfig {
+        // Request URI that has the JWKs.
+        // Required.
+        string jwks_uri = 1;
+        
+        // Request interval to check whether new JWKs are available.
+        // Required.
+        int32 request_duration_sec = 2;
```
incfly, line 87: name to `periodic_fetch_interval`?  request_duration sounds like how long a single HTTP request can last.

config/oidc/config.proto
```diff
@@ -75,14 +75,30 @@ message OIDCConfig {
...
+        // The JSON JWKS response from the OIDC provider’s `jwks_uri` URI which can be found in
+        // the OIDC provider's
+        // [configuration response](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse).
+        // Note that this JSON value must be escaped when embedded in a json configmap
+        // (see [example](https://github.com/istio-ecosystem/authservice/blob/master/bookinfo-example/config/authservice-configmap-template.yaml)).
+        // Used during token verification.
+        string jwks = 4;
+
+        // Configuration to allow JWKs to be retrieved and updated asynchronously at regular intervals.
+        JwksFetcherConfig jwks_fetcher = 17;
```
incfly, line 100: should we add a NEXT tag in the beginning of the proto to reminds future editing what's next available number? smt like https://github.com/envoyproxy/envoy/blob/main/api/envoy/data/cluster/v2alpha/outlier_detection_event.proto#L77

src/filters/filter_chain.cc
```diff
@@ -16,16 +19,32 @@
...
+    if (filter.oidc().jwks_config_case() == config::oidc::OIDCConfig::kJwks) {
+      jwks_storage_map_.emplace_back(
+          std::make_shared<oidc::PermanentJwksStorageImpl>(
+              filter.oidc().jwks(), google::jwt_verify::Jwks::Type::JWKS));
+    } else {
+      jwks_storage_map_.emplace_back(
+          std::make_shared<oidc::NonPermanentJwksStorageImpl>(
+              filter.oidc().jwks_fetcher().jwks_uri(),
+              std::chrono::seconds(
+                  filter.oidc().jwks_fetcher().request_duration_sec()),
```
incfly, line 38: is the default value is 0 if not specified? shall we have some sanity check to make a minimum, say 600 seconds?

for reference, istiod default jwks update is 20min. https://istio.io/latest/docs/reference/commands/pilot-discovery/. search for "PILOT_JWT_PUB_KEY_REFRESH_INTERVAL".

src/filters/oidc/jwks_storage.h
```diff
@@ -0,0 +1,111 @@
...
+  virtual google::jwt_verify::JwksPtr jwks() override {
+    return parseJwks(jwks_, type_);
+  }
+
+ private:
+  std::string jwks_;
+  google::jwt_verify::Jwks::Type type_;
+};
+
+class NonPermanentJwksStorageImpl : public JwksStorage {
```
incfly, line 59: I feel the NonPermanent is a bit confusing, it reads to me that it can be volatile. Whereas the main benefits is based on jwks uri and fetched dynamically. 

I think these two classes if named as "StaticJwksResolver" and "DynamicJwksResolver" will be easier to follow. WDTY?

test/filters/oidc/jwks_storage_test.cc
```diff
@@ -0,0 +1,69 @@
+#include "src/filters/oidc/jwks_storage.h"
+
+#include "gtest/gtest.h"
+
+namespace authservice {
+namespace filters {
+namespace oidc {
+
+namespace {
+// A good public key based on above private key
```
incfly, line 10: not sure i follow this comments. where is the "above private key"? this is at the top of the file.

incfly, line 10:
```suggest
// A valid public JWK key for JWT verification.
const char valid_jwks_public_key_ [] = R"(
```

incfly: signing key means private key in these contexts. I think we are testing public key part.

test/filters/oidc/jwks_storage_test.cc
```diff
@@ -0,0 +1,69 @@
...
+const char valid_jwt_signing_key_[] = R"(
```
incfly, line 11: same here.

test/filters/oidc/jwks_storage_test.cc
```diff
@@ -0,0 +1,69 @@
...
+      "kid": "b3319a147514df7ee5e4bcdee51350cc890cc89e",
+      "n":
+      "up97uqrF9MWOPaPkwSaBeuAPLOr9FKcaWGdVEGzQ4f3Zq5WKVZowx9TCBxmImNJ1qmUi13pB8otwM_l5lfY1AFBMxVbQCUXntLovhDaiSvYp4wGDjFzQiYA-pUq8h6MUZBnhleYrkU7XlCBwNVyN8qNMkpLA7KFZYz-486GnV2NIJJx_4BGa3HdKwQGxi2tjuQsQvao5W4xmSVaaEWopBwMy2QmlhSFQuPUpTaywTqUcUq_6SfAHhZ4IDa_FxEd2c2z8gFGtfst9cY3lRYf-c_ZdboY3mqN9Su3-j3z5r2SHWlhB_LNAjyWlBGsvbGPlTqDziYQwZN4aGsqVKQb9Vw",
+      "e": "AQAB"
+    }
+  ]
+}
+)";
+
+const char *invalid_jwt_signing_key_ =
```
incfly, line 36: just curious, what doe this base64 value stand for? if it's a dummy short base64 encoded "foo", or literal "foo", the test should work as well, right?

src/filters/oidc/jwks_storage.cc
```diff
@@ -0,0 +1,38 @@
...
+      [this](const boost::system::error_code& ec) { this->request(ec); });
+}
+
+void NonPermanentJwksStorageImpl::JwksFetcher::request(
+    const boost::system::error_code&) {
+  boost::asio::spawn(ioc_, [this](boost::asio::yield_context yield) {
+    auto resp = http_ptr_->Get(jwks_uri_, {}, "", "", "", ioc_, yield);
+    const auto& fetched_jwks = resp->body();
+    // TODO(shikugawa): Prevent lock with same jwks.
+    // TODO(shikugawa): Support PEM.
```
incfly, line 27: not sure if understand. what's the PEM used in this case? I am not aware of any keys that put PEM values in JWK or you mean something else? any link for the reference/example?

src/filters/oidc/jwks_storage.h
```diff
@@ -0,0 +1,111 @@
...
+
+  google::jwt_verify::JwksPtr jwks() override {
+    std::unique_lock<std::mutex> lck(mux_);
+    return parseJwks(jwks_, type_);
+  }
+
+ private:
+  std::string jwks_;
+  google::jwt_verify::Jwks::Type type_;
+  std::unique_ptr<JwksFetcher> jwks_fetcher_;
```
incfly, line 103: it'll be good if we can have macro to annotate which fields are protected under the mutex. ABSL mutex should support that. not sure about std library. @lizan might know more.

test/filters/oidc/jwks_storage_test.cc
```diff
@@ -0,0 +1,69 @@
...
+  EXPECT_EQ(google::jwt_verify::Status::Ok, storage.jwks()->getStatus());
+
+  PermanentJwksStorageImpl storage2(invalid_jwt_signing_key_,
+                                    google::jwt_verify::Jwks::JWKS);
+  EXPECT_NE(google::jwt_verify::Status::Ok, storage2.jwks()->getStatus());
+}
+
+TEST(JwksStorageTest, TestNonPermanentJwksStorage) {
+  boost::asio::io_context io_context;
+  NonPermanentJwksStorageImpl storage("istio.io", std::chrono::seconds(0),
```
incfly, line 56: so this is not actually testing the actual periodic functionality right? can we test that behavior? kick off the `request()` and wait enough to get the updates, instead of calling updateJwks` directly?

Since that's what the caller/user of this libary expects.

src/filters/oidc/jwks_storage.h
Shikugawa: Yeah that makes sense for me.

test/filters/oidc/jwks_storage_test.cc
Shikugawa: in this test, we should have to check that JwksStorage can resolve internal JWKs or update them. So this test can accept no matter what `invalid_jwt_signing_key` is, I think.

src/filters/oidc/jwks_storage.cc
Shikugawa: This is not needed here. delete them.

config/oidc/config.proto
Shikugawa: These free field is generated automatically on envoy with complex protoxform scripting. But it wiil be difficult to automate them in this tiny repository. But it will be good to have manually. WDYT?

test/filters/oidc/jwks_storage_test.cc
Shikugawa: Added realistic test here.

Shikugawa:
@incfly Our requirements which what you said in the discussion, 

> There's no call to validate the parse results of JWKS at the filter constructing time. the error only occurs at the runtime during oidc flow, but we don't give any error message.

It may be accomplished with this PR simultaneously because we changed to construct `StaticJwksResolver` at `FilterChain` constructing time.

PR #153, opened 2021-08-20
Update a few issues w.r.t jwk update and listeners.

### LINKED ISSUE #149: Update the authservice to the Istio 1.9+
The doc and the setup requires some setup. This issue captures the changes including

- ExternalAuthz
- UX improvements, etc localhost and port forwarding for quicker "get started".
- Doc structure improvement for different deployment model.
- Purge the outdated content.

incfly:
Since 1.10 Istio no longer support app listening on port localhost.
Also update the doc of how to set up jwk, which could be updated.

Fix #149 

Signed-off-by: Jianfei Hu <jianfeih@tetrate.io>
Signed-off-by: Jianfei Hu <jianfeih@google.com>

### DIFF
--- a/bookinfo-example/README.md
+++ b/bookinfo-example/README.md
@@ -52,6 +52,18 @@ URI to be hosted on a protected endpoint.
            port: "10003"
    ```
 
+1. Fetch the identity provider public key and populate into the configmap. In our example, run
+`scripts/google-jwks.sh`.
+
+      ```shell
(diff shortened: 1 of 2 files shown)


incfly:
tried to plumb through helm values for the jwk content a bit. not work well since it has some string escape issues (when using --set from helm cli, the escaped value is unescaped...)

suboptimial but at least offer a way. we won't need this once jwkURI fectching is supported..

PR #160, opened 2021-09-23
fix invalid loglevel on OIDC filter

Shikugawa:


test/filters/oidc/jwks_resolver_test.cc
```diff
@@ -66,10 +66,10 @@ const char invalid_jwt_public_key_[] = R"(
         {
            "kty": "XYZ",
            "crv": "P-256",
-           "x": "unko",
-           "y": "unko",
-           "alg": "unko",
-           "kid": "fuck"
```
incfly, line 72: I learned what "unko" means just now 😄

src/filters/oidc/token_response.cc
```diff
@@ -81,7 +81,7 @@ std::shared_ptr<TokenResponse> TokenResponseParserImpl::Parse(
   const auto status = ::google::protobuf::util::JsonStringToMessage(
       raw_string_piece, &message, options);
   if (!status.ok()) {
-    spdlog::info("{}: JSON parsing error: {}", __func__,
+    spdlog::warn("{}: JSON parsing error: {}", __func__,
```
incfly, line 84: why not `error` and for the rest?

test/filters/oidc/jwks_resolver_test.cc
Shikugawa: 🤷‍♂

src/filters/oidc/token_response.cc
Shikugawa: My consideration is just "system error" should be `error`. But here is not a system error.

Shikugawa: My consideration is just severe "system error" that won't be processed anymore should be `error`. But here is not so critical, and no problem to continue entire process.

Shikugawa:
@incfly @lizan Could you approve this?

incfly (approved):

Issue #34, opened 2019-10-14
Dynamically fetch jwks based on configured URI


tylerschultz:
Placeholder issue. Details TBD.

nickrmc83:
See https://openid.net/specs/openid-connect-discovery-1_0.html

sebster:
This would be really useful to have. Especially with many environments and many back-ends, it makes the configuration less error prone and a lot more convenient.

incfly:
Seems like we already have some feature requests. @Shikugawa is already working on this.

incfly:
@Shikugawa Can you work on providing documentation and verify this works E2E? I tried myself but not working when using JWKS URI, (changing this back to the inlined jwks would work again).

---

## Ways to reproduce

1. Follow instructions in book-info/readme.md
2. Setup image to  (You can build your own and swap to make sure), via helm --set authservie.image="ghcr.io/istio-ecosystem/authservice/authservice:dev"
3. Change the [config.yaml](https://github.com/istio-ecosystem/authservice/blob/f7c9fca91007758bbbf9b5a285dcab5d61c636b7/bookinfo-example/authservice/templates/config.yaml#L33) as 

```yaml
"jwks_fetcher": {
                  "jwks_uri": "https://www.googleapis.com/oauth2/v3/certs"
                },
```
4. port forwarding, send requests, etc. You will see the request stuck at the callback. the authserivce pod has the error as 

```
authservice-599f996476-csp9k [2021-09-20 21:50:54.923[] [console[] [debug[] Check: processing request https://localhost:8443/productpage/oauth/callback?state=-xfxptk_CDrPWVwfi8z1hZo-V7QqbqY-cJQ-onbsCp4&code=4%2F0AX4XfWjIne-ddSme9vAwit5uHWq1L9iiTMqrZkMOJtLFOn_cX65gql3OsSOgpuWCyDJIKQ&scope=openid&authuser=0&prompt=none with filter chain idp_filter_chain                                                                                                                                                                authservice-599f996476-csp9k [2021-09-20 21:50:54.923[] [console[] [trace[] New
authservice-599f996476-csp9k [2021-09-20 21:50:54.923[] [console[] [trace[] OidcFilter
authservice-599f996476-csp9k [2021-09-20 21:50:54.923[] [console[] [trace[] Process
authservice-599f996476-csp9k [2021-09-20 21:50:54.923[] [console[] [debug[] Call from @127.0.0.1 to CN=localhost@127.0.0.1
[...]

PR #161, opened 2021-09-23
enable to fetch JWKs on dynamic JwkResolver after system started

Shikugawa:
Fix a bug described here. https://github.com/istio-ecosystem/authservice/issues/34#issuecomment-923350828
The cause of this bug is not to be invoked immediately after the system started. The default timer is 20min so that JWKs will be empty after 20min elapsed after the system started in the previous implementation. This PR fixes this unexpected behavior.


src/filters/oidc/jwks_resolver.cc
```diff
@@ -15,20 +15,23 @@ DynamicJwksResolverImpl::JwksFetcher::JwksFetcher(
...
     auto resp = http_ptr_->Get(jwks_uri_, {}, "", "", "", ioc_, yield);
+    auto next_schedule_interval = periodic_fetch_interval_sec_;
 
     if (resp == nullptr) {
-      spdlog::info("{}: HTTP connection error", __func__);
+      spdlog::warn("{}: HTTP connection error", __func__);
     } else {
       if (resp->result() != boost::beast::http::status::ok) {
-        spdlog::info("{}: HTTP response error: {}", __func__,
+        spdlog::warn("{}: HTTP response error: {}", __func__,
```
incfly, line 34: this can be "error" level?

src/filters/oidc/jwks_resolver.cc
```diff
@@ -37,8 +40,19 @@ void DynamicJwksResolverImpl::JwksFetcher::request(
...
 
+    if (parent_->jwks() == nullptr) {
+      // On Kubernetes, depending on the timing of the Pod startup,
+      // the first egress HTTP/HTTPS request may fail. This problem
+      // goes away after a few seconds from the start of the pod,
+      // but by default, the timer is rescheduled after 20 minutes,
+      // so if the first fetch fails, the JWKs will be empty for 20 minutes.
+      // To prevent this problem, the timer should be rescheduled after 3
+      // seconds if the acquisition of JWKs fails.
+      next_schedule_interval = std::chrono::seconds(3);
```
incfly, line 51: this this makes sure that we always retry every 3 seconds util the first jwks is successsfully initialized, is that correct?

incfly (commented):
Can you add a unit test in https://github.com/istio-ecosystem/authservice/blob/master/test/filters/oidc/jwks_resolver_test.cc? Basically simulate that the first 5 seconds, the mock http server always returns failure. then this PR should kick in. (and the test would fail without this PR)

src/filters/oidc/jwks_resolver.cc
Shikugawa: Yes. Once JWKs is set, it will never be set to true. This means that it will only be called if a connection error or HTTP error occurs and JWKs are not set from the system startup.

src/filters/oidc/jwks_resolver.cc
Shikugawa: The reason is same as https://github.com/istio-ecosystem/authservice/pull/160#discussion_r715548073

test/filters/oidc/jwks_resolver_test.cc
```diff
@@ -155,6 +155,39 @@ TEST(JwksResolverTest, TestDynamicJwksResolverWithInvalidHttpStatus) {
   EXPECT_EQ(nullptr, resolver.jwks());
 }
 
+TEST(JwksResolverTest,
+     TestDynamicJwksResolver_RequestBy3SecInterval_UntilJwksConfigured) {
```
incfly, line 159: The other test name is CamelCase, we should be consistent and not use underscore.

src/filters/oidc/jwks_resolver.cc
```diff
@@ -37,8 +42,22 @@ void DynamicJwksResolverImpl::JwksFetcher::request(
       }
     }
 
+    constexpr auto default_interval_with_no_jwks = std::chrono::seconds(3);
```
incfly, line 45: following envoy code practice, if I remember correctly, constants are usually defined outside the function, starting with letter `kXXX`.

incfly: also feel the var can be named a bit different for readability.

```cpp
// The number of seconds dynamic jwks resolver will retry to request JWKS during the startup time.
// A short interval ensures us can get JWKS ready before serving the traffic.
constexpr int kRefreshSecondsBeforeInitialized = 3;
```

test/filters/oidc/jwks_resolver_test.cc
```diff
@@ -155,6 +155,39 @@ TEST(JwksResolverTest, TestDynamicJwksResolverWithInvalidHttpStatus) {
...
+                    const std::map<absl::string_view, absl::string_view>,
+                    absl::string_view, absl::string_view, absl::string_view,
+                    boost::asio::io_context&, boost::asio::yield_context) {
+            common::http::response_t response = std::make_unique<
+                beast::http::response<beast::http::string_body>>();
+            response->result(503);
+            return response;
+          }));
+
+  io_context.run_for(std::chrono::seconds(3));
```
incfly, line 181: We should sleep slightly longer than interval right? This ensures the retries would be kicked in. say 4 seconds.

test/filters/oidc/jwks_resolver_test.cc
```diff
@@ -155,6 +155,39 @@ TEST(JwksResolverTest, TestDynamicJwksResolverWithInvalidHttpStatus) {
...
+     TestDynamicJwksResolver_RequestBy3SecInterval_UntilJwksConfigured) {
+  boost::asio::io_context io_context;
+  auto mock_http = std::make_shared<common::http::HttpMock>();
+
+  // Configured request interval as 10000 sec. This value is enough to guarantee
+  // that second request will not be invoked for 3 sec evloop run.
+  DynamicJwksResolverImpl resolver("istio.io", std::chrono::seconds(10000),
+                                   mock_http, io_context);
+
+  // Never initialized with invalid HTTP status.
```
incfly, line 168: If I understand correctly, we want to make the initial setup always return failure to check jwks as nullptr.

```
// Initially make mock server always return invalid HTTP response as 503.
```

src/filters/oidc/jwks_resolver.cc
```diff
@@ -37,8 +42,22 @@ void DynamicJwksResolverImpl::JwksFetcher::request(
...
+
+    if (next_schedule_interval > default_interval_with_no_jwks &&
+        parent_->jwks() == nullptr) {
+      // On Kubernetes, depending on the timing of the Pod startup,
```
incfly, line 49: So if I understand correctly, the following error case might still happen:

1. pod comes up.
2. k8s network glitches, jwks request fails for whatever resaon ( could also be due to IdP is temporarily not available)
3. External Authz Check request comes in. However at this point we still don't have jwks ready yet for JWT verification.

Is that right? Ideally we should make the jwks fetching completion as the criteria for the authservice pod [readinessProbe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#http-probes)

if above is correct, can we create a TODO and log a bug to track this? Don't have to be done in this PR for sure.

incfly:
sorry somehow this slipped my task list. just took another pass.

src/filters/oidc/jwks_resolver.cc
Shikugawa: I think that it can be better to have the feature to configure `initialDelaySeconds` that acts to start the initial request after these seconds. I will change the implementation of this PR.

Also, The idea seems great. My idea is to have `/health` endpoint that returns 200 not only authservice is fine but JWKs are ready.

Shikugawa:
@incfly @lizan Could you approve this?

src/filters/oidc/jwks_resolver.cc
incfly: I think given you have the /health endpoint, and prober logic will prevent us from resulting into the above failure case, right? I would rather keep it that way, since how long to delay, (3seconds), whether to delay is really up to the environmental issue.

A reconciling model of probing /health endpoint would ensure us to eventual get there with less knobs. Customers would also not have to configure one more field. Trying to understand the impact of different values to the startup success.

docs/README.md
```diff
@@ -88,6 +88,10 @@ The configuration of an OpenID Connect filter that can be used to retrieve ident
 | JwksFetcherConfig | This message defines a setting to allow asynchronous retrieval and update of the JWK for JWT validation at regular intervals. | message |
 | jwks_uri | Request URI that has the JWKs. Required. | string |
 | periodic_fetch_interval_sec | Request interval to check whether new JWKs are available. If not specified, default to 1200 seconds, 20min. Optional. | uint32 |
+<<<<<<< HEAD
```
incfly, line 91: here's a merge conflicit seems like.

incfly (commented):
sorry for the delay. Only one nitpick, seeing since you already have /health endpoint, I feel we don't need the `initial_fetch_delay_sec` as an option. (See comments below)b

PR #164, opened 2021-09-30
implement healthcheck endpoint for active JWKs existence

Shikugawa:


incfly:
seems like this includes the abstraction refactoring changes you have in another PR? Maybe we can land that one first so that the diff here is more obvious.

Shikugawa:
Yes. This PR needs abstraction to achieve DI of JWK resolver for test. I will start to add new test after that landed.

PR #165, opened 2021-10-05
filter and session store abstraction and filter chain refactor

Shikugawa:

### DIFF
--- a/src/filters/filter_chain.h
+++ b/src/filters/filter_chain.h
@@ -2,12 +2,14 @@
 #define AUTHSERVICE_FILTER_CHAIN_H
 
 #include <memory>
+#include <vector>
 
 #include "boost/asio/io_context.hpp"
 #include "config/config.pb.h"
 #include "config/oidc/config.pb.h"
 #include "envoy/service/auth/v3/external_auth.grpc.pb.h"
 #include "src/filters/filter.h"
+#include "src/filters/filter_factory.h"
 #include "src/filters/oidc/jwks_resolver.h"
 #include "src/filters/oidc/session_store.h"
 
@@ -54,8 +56,9 @@ class FilterChainImpl : public FilterChain {
  private:
   unsigned int threads_;
   config::FilterChain config_;
-  std::shared_ptr<oidc::SessionStore> oidc_session_store_;
-  std::vector<std::shared_ptr<oidc::JwksResolver>> jwks_resolver_map_;
+  oidc::SessionStorePtr oidc_session_store_;
+  oidc::JwksResolverCachePtr jwks_resolver_cache_;
+  std::vector<FilterFactoryPtr> filter_factory_chain_;
 
  public:
   explicit FilterChainImpl(boost::asio::io_context &ioc,
--- a/src/filters/filter_chain.h
+++ b/src/filters/filter_chain.h
@@ -58,7 +58,7 @@ class FilterChainImpl : public FilterChain {
   config::FilterChain config_;
   oidc::SessionStorePtr oidc_session_store_;
   oidc::JwksResolverCachePtr jwks_resolver_cache_;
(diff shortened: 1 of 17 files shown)


src/filters/filter_chain.h
```diff
@@ -54,8 +56,9 @@ class FilterChainImpl : public FilterChain {
  private:
   unsigned int threads_;
   config::FilterChain config_;
-  std::shared_ptr<oidc::SessionStore> oidc_session_store_;
-  std::vector<std::shared_ptr<oidc::JwksResolver>> jwks_resolver_map_;
+  oidc::SessionStorePtr oidc_session_store_;
+  oidc::JwksResolverCachePtr jwks_resolver_cache_;
+  std::vector<FilterFactoryPtr> filter_factory_chain_;
```
incfly, line 61: to its semantics, this would be better named as `filter_factories_` right? Chain has its own meaning and here can have audience think it's related to the filter chain. Simple plural form should be enough.

src/filters/filter_factory.h
```diff
@@ -0,0 +1,26 @@
...
+
+namespace authservice {
+namespace filters {
+
+class FilterFactory {
+ public:
+  virtual ~FilterFactory() = default;
+
+  /**
+   * Create auth filter in filter chain.
```
incfly, line 16:
```suggestion
Creates an authentication filter.
```

incfly: We don't have to mention where the filter is used. that's up to the caller.

src/filters/oidc/redis_session_store.h
```diff
@@ -53,6 +55,25 @@ class RedisSessionStore : public SessionStore {
...
+      : threads_(threads),
+        absolute_session_timeout_(absolute_session_timeout),
+        idle_session_timeout_(idle_session_timeout),
+        config_(config) {}
+
+  SessionStorePtr create() override;
+
+ private:
+  const int threads_ = 1;
+  uint32_t absolute_session_timeout_ = 0;
```
incfly, line 72: does these absolute_ fields also make sense to be constant?

src/filters/oidc/jwks_resolver.h
```diff
@@ -109,6 +113,43 @@ class DynamicJwksResolverImpl : public JwksResolver {
...
+    switch (config_.jwks_config_case()) {
+      case config::oidc::OIDCConfig::kJwks:
+        resolver_ =
+            std::make_shared<oidc::StaticJwksResolverImpl>(config_.jwks());
+        break;
+      case config::oidc::OIDCConfig::kJwksFetcher: {
+        uint32_t periodic_fetch_interval_sec =
+            config_.jwks_fetcher().periodic_fetch_interval_sec();
+        if (periodic_fetch_interval_sec == 0) {
+          periodic_fetch_interval_sec = 1200;
```
incfly, line 130: make this a constant var like `kJwksPeriodicFetchIntervalSec`, versus in-place numeric value?

src/filters/oidc/in_memory_session_store.h
```diff
@@ -52,6 +53,20 @@ class InMemorySessionStore : public SessionStore {
...
+ public:
+  InMemorySessionStoreFactory(uint32_t absolute_session_timeout,
+                              uint32_t idle_session_timeout)
+      : absolute_session_timeout_(absolute_session_timeout),
+        idle_session_timeout_(idle_session_timeout) {}
+
+  SessionStorePtr create() override;
+
+ private:
+  uint32_t absolute_session_timeout_ = 0;
```
incfly, line 66: same here. will these fields change?

incfly (commented):
LG overall. Thanks for the change.

incfly (approved):

PR #167, opened 2021-10-08
fix OOM while processing gRPC completion state

Shikugawa:
The reason for the OOM was that the ProcessingState pointer passed to CompletionState was not released properly. As a result of the fix, the memory pressure is as follows The load test was done as follows for bookinfo deployed in GKE.

### before
```
fortio load -H ${COOKIE} -qps 800 -t 6m -c 4 https://${AUTHORITY_NAME}/productpage
```

Screenshot from 2021-10-08 18-02-19

### after
```
fortio load -H ${COOKIE} -qps 6000 -t 6m -c 4 https://${AUTHORITY_NAME}/productpage
```

Screenshot from 2021-10-08 18-02-07

### DIFF
--- a/src/service/async_service_impl.cc
+++ b/src/service/async_service_impl.cc
@@ -86,10 +86,8 @@ void ProcessingState::Proceed() {
 void CompleteState::Proceed() {
   spdlog::trace("Processing completion and deleting state");
 
-  if (!processor_v2_) delete processor_v2_;
-
-  if (!processor_v3_) delete processor_v3_;
-
+  delete processor_v2_;
+  delete processor_v3_;
   delete this;
 }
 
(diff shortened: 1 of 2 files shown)


incfly:
I read the gRPC doc for the async C++ impl. https://grpc.io/docs/languages/cpp/async/
And here is the sample server, with their `Process` implementation

https://github.com/grpc/grpc/blob/v1.41.0/examples/cpp/helloworld/greeter_async_server.cc#L86

You can see official example handle the event as `CREATE, PROCESS, FINISH`, and only create a new instance of the processing state with `new` when status is `PROCESS` branch. Because at that point, the current instance of the state is assigned to handle the current request.

However, in our code, processingstate::process is always invoking `parent_.create()` to new an instance. This means when the state is either `CREATE` or `PROCESS` we will create a new instance of process.

Compared with official e.g, we create more instances.

Therefore I think that's the root cause of the OOM.

Shikugawa:
Sorry. The graph above has the wrong point. `(1) processing state V2` that created at the initial phase won't be released after `completion_state_->Proceed()` called. (But It is not related with OOM of course)

PR #168, opened 2021-10-08
fix potential I/O thread stall with non-termination of event queue with periodic task

Shikugawa:
In past implementations, the periodic task, and the rescheduling of the periodic task, would not terminate even though the Graceful Shutdown of the server was terminated, so the I/O thread would not terminate forever. By forcing the event loop to terminate with this change, you can safely wait for the I/O thread to terminate regardless of whether or not the periodic task is in the event queue.

cc: @cfryanr


Shikugawa:
@incfly Could we proceed this? this is needed to prevent potentially unexpected behavior that will cause I/O thread stall.

src/service/async_service_impl.cc
```diff
@@ -221,6 +221,11 @@ void AsyncAuthServiceImpl::Run() {
   // Reset the work item for the IO service will terminate once it finishes any
   // outstanding jobs
   work.reset();
+
+  // This is used to prevent schedule next periodic cleanup task.
```
incfly, line 225:
```suggestion
  // The `SchedulePeriodicCleanupTask` is scheduled with `timer_`, which is bound with the io_context_.
  // Calling `stop()` explicitly ensures that the periodic task won't be further scheduled onto `io_context`
  // internal scheduling queue. Therefore the thread running periodic task can be terminated and joined here.
```

incfly (commented):
Thanks for the detailed explaining, just one nitpick for the comment rewording.

Another optional thing, not particular related to this PR, I feel the line 202 - 224 can be wrapped into a separate `Stop/Shutdown` method by itself. `Run` itself is already long enough.

PR #170, opened 2021-10-13
prevent invalid JWKs in StaticJwksResolver

Shikugawa:

In previous implementation, invalid JWKs will be accepted in `StaticJwksResolver`. With this PR, it will throw exception if invalid JWKs are passed through configuration.

### DIFF
--- a/test/config/getconfig_test.cc
+++ b/test/config/getconfig_test.cc
@@ -137,29 +137,28 @@ TEST_F(GetConfigTest, ReturnsTheConfig) {
   ASSERT_EQ(oidc.authorization_uri(), "https://google3/path3");
   ASSERT_EQ(oidc.token_uri(), "https://google2/path2");
   ASSERT_EQ(oidc.callback_uri(), "https://google4/path4");
-  ASSERT_EQ(
-      oidc.jwks(),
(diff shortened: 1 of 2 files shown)


test/config/getconfig_test.cc
```diff
@@ -137,7 +137,29 @@ TEST_F(GetConfigTest, ReturnsTheConfig) {
   ASSERT_EQ(oidc.authorization_uri(), "https://google3/path3");
   ASSERT_EQ(oidc.token_uri(), "https://google2/path2");
   ASSERT_EQ(oidc.callback_uri(), "https://google4/path4");
-  ASSERT_EQ(oidc.jwks(), "jwks_placeholder");
+  ASSERT_EQ(
+      oidc.jwks(),
+      "{\n  \"keys\": [\n    {\n      \"e\": \"AQAB\",\n      \"kty\": "
```
incfly, line 142: you can use multi line string literal for easier editing.
https://stackoverflow.com/a/5460235

Shikugawa:
> thanks for the change. So if the users are setting up the invalid static jwks, what are the behavior different before and after this PR? Before I think authservice will start successfully but fails as the jwt id token verification. how will it look like after this PR?

Yes. In previous code, if the user passed invalid JWKs then failed to verify token. But it is very odd behavior. We should assert users the JWKs are invalid.

PR #174, opened 2021-10-15
implement healthcheck endpoint for active JWKs existence

Shikugawa:


incfly (commented):
This PR makes it as a gRPC service for the checking. Is that necessary? Kubernetes supports HTTP probes only. https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/ 

I thought we can open a port and serve the health check status there. And the k8s deployment YAML will specify the health check roughly like

```yaml
readinessProbe:
  httpGet:
    path: /healthz
    port: 8080 # or something else.
```

src/service/async_service_impl.h
```diff
@@ -263,7 +228,6 @@ class AsyncAuthServiceImpl {
 
   envoy::service::auth::v2::Authorization::AsyncService service_v2_;
   envoy::service::auth::v3::Authorization::AsyncService service_;
-  grpc::health::v1::Health::AsyncService health_service_;
```
incfly: Instead of  calling `new HealthCheckServer` in the `Run` directly, can we keep making the server as a field owned by the AsyncAuthserviceImpl? This way it's clearer the class has the health check functionality and object ownership (as opposed to new).

src/service/healthcheck_http_server.h
```diff
@@ -0,0 +1,115 @@
...
+  void onReadDone() {
+    response_.version(request_.version());
+
+    http::status status = http::status::ok;
+
+    if (request_.method() != http::verb::get ||
+        request_.target() != "/healthz") {
+      status = http::status::bad_request;
+    } else {
+      for (auto &&chain : chains_) {
```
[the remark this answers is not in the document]
incfly: including which chains (using index or better identifier like the match/name).

bookinfo-example/config/authservice-configmap-template-for-authn-and-authz.yaml
```diff
@@ -15,6 +15,7 @@ data:
     {
       "listen_address": "127.0.0.1",
       "listen_port": "10003",
+      "healthz_listen_port": "9999",
```
incfly, line 18: I think it'll be better to default the port to 10004, next to the serving port.

This allows a easier firewall configuration in real production environment. Security sensitive people only allow a workload to listen on a list of port. It's more straighforward to be a narrow and consecutive range.

src/service/healthcheck_http_server.h
```diff
@@ -0,0 +1,115 @@
...
+namespace authservice {
+namespace service {
+
+namespace beast = boost::beast;    // from <boost/beast.hpp>
+namespace http = beast::http;      // from <boost/beast/http.hpp>
+using tcp = boost::asio::ip::tcp;  // from <boost/asio/ip/tcp.hpp>
+
+class HealthcheckHttpConnection {
+ public:
+  HealthcheckHttpConnection(
```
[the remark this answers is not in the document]
Shikugawa: added

src/service/healthcheck_http_server.h
```diff
@@ -0,0 +1,115 @@
...
+#include "src/filters/filter_chain.h"
+
+namespace authservice {
+namespace service {
+
+namespace beast = boost::beast;    // from <boost/beast.hpp>
+namespace http = beast::http;      // from <boost/beast/http.hpp>
+using tcp = boost::asio::ip::tcp;  // from <boost/asio/ip/tcp.hpp>
+
+class HealthcheckHttpConnection {
```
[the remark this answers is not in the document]
Shikugawa: added

src/common/http/http.cc
```diff
@@ -621,6 +623,71 @@ response_t HttpImpl::Get(
   }
 }
 
+response_t HttpImpl::SimpleGet(
```
Shikugawa, line 626: We don't have transport socket abstraction (raw or tls). This is why we must add another redundant simple HTTP request function. We should introduce transport socket abstraction to clean up these codes.

config/config.proto
```diff
@@ -121,6 +121,14 @@ message Config {
...
     // Optional.
     oidc.OIDCConfig default_oidc_config = 10;
+
+    // The TCP port for the Authservice to listen for incorming REST request for healthcheck.
+    // Healthcheck is used to check not only for checking that server is up and running
+    // but valid JWKs for token validation are extracted and available for request validation.
+    // This feature is intended to be used from [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
+    // on Kubernetes.
+    // Required.
+    int32 healthz_listen_port = 11 [(validate.rules).int32.lt = 65536];
```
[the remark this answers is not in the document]
Shikugawa: deleted for now

bookinfo-example/authservice/templates/deployment.yaml
```diff
@@ -29,6 +29,9 @@ spec:
         volumeMounts:
           - name: authservice-config # mount the volume containing the authservice ConfigMap
             mountPath: /etc/authservice
+        livenessProbe:
```
incfly, line 32: I think we should use readinessProbe. Liveness probe will continuously be checked by k8s and if fails, pod is killed. however, we only care about the pod startup time, authsvc is able to fetch the jwks public key.

src/common/http/http.cc
```diff
@@ -621,6 +622,71 @@ response_t HttpImpl::Get(
...
+    tcp::resolver resolver(ioc);
+    beast::tcp_stream stream(ioc);
+
+    spdlog::info("{}: opening connection to {}:{}", __func__,
+                 parsed_uri.GetHost(), parsed_uri.GetPort());
+    const auto results = resolver.async_resolve(
+        parsed_uri.GetHost(), std::to_string(parsed_uri.GetPort()), yield);
+    beast::get_lowest_layer(stream).async_connect(results, yield);
+
+    // Set up an HTTP POST request message
```
incfly, line 645: the method is named SimpleHTTPGet. the code below is also get. line 647.

src/common/http/http.cc
```diff
@@ -621,6 +622,71 @@ response_t HttpImpl::Get(
...
+    beast::flat_buffer buffer;
+    response_t res(new beast::http::response<beast::http::string_body>);
+    beast::http::async_read(stream, buffer, *res, yield);
+
+    // Gracefully close the socket.
+    // Receive an error code instead of throwing an exception if this fails, so
+    // we can ignore some expected not_connected errors.
+    boost::system::error_code ec;
+
+    if (ec) {
```
incfly, line 670: what's the purpose of this check, ec is just created in line 668 without anything modifying it, are we checking the default value?

src/common/http/http.cc
```diff
@@ -621,6 +622,71 @@ response_t HttpImpl::Get(
...
+        spdlog::error("{}: HTTP error encountered: {}", __func__, ec.message());
+        return response_t();
+      }
+    }
+
+    return res;
+    // If we get here then the connection is closed gracefully
+  } catch (std::exception const &e) {
+    spdlog::error("{}: unexpected exception: {}", __func__, e.what());
+    return response_t();
```
incfly, line 686: this will return a default response_t to the caller. and the exception is handled within the simpleGet. how do we expect the caller to check whether return value is legit or not? absl::StatusOr seems a good fit for this.

src/service/healthcheck_http_server.cc
```diff
@@ -0,0 +1,86 @@
...
+void HealthcheckHttpConnection::onReadDone() {
+  response_.version(request_.version());
+  http::status status = http::status::ok;
+
+  if (request_.method() != http::verb::get || request_.target() != "/healthz") {
+    status = http::status::bad_request;
+  } else {
+    for (auto&& chain : chains_) {
+      if (!chain->jwksActive()) {
+        spdlog::warn("{}: chain:{} is inactive yet", __func__, chain->Name());
```
incfly, line 35: make it clear that chain:{} JWKS is not ready yet? people who read the log will not see the code around here immediately and adding jwks help them for a faster troubleshooting time.

test/service/healthcheck_http_server_test.cc
```diff
@@ -0,0 +1,124 @@
...
+      .Times(2)
+      .WillRepeatedly(Return(mock_resolver));
+
+  chain->setJwksResolverCache(std::move(resolver_cache));
+  EXPECT_FALSE(chain->jwksActive());
+
+  std::vector<std::unique_ptr<filters::FilterChain>> chains;
+  chains.push_back(std::move(chain));
+
+  HealthcheckAsyncServer server(chains, "0.0.0.0", 33333);
```
incfly, line 47: There'll be small chance the port is not ready when CI/test runs. What I see oten unit test usually do is using "0" port and get the allocated port from system.

test/service/healthcheck_http_server_test.cc
```diff
@@ -0,0 +1,124 @@
...
+
+  auto http_ptr = common::http::ptr_t(new common::http::HttpImpl);
```
incfly, line 49: It takes me a bit browsing to realize this is a shared_ptr. Is this the common typedef naming pattern we use in other places as well?

Also I think here unique_ptr would be sufficient right? And unique_ptr is preferred over shared_ptr when possible IIRC.

src/service/healthcheck_http_server.cc
```diff
@@ -0,0 +1,86 @@
...
+HealthcheckAsyncServer::~HealthcheckAsyncServer() {
+  for (auto&& conn : active_connections_) {
+    delete conn;
+  }
+  ioc_.stop();
+  th_.join();
+}
+
+void HealthcheckAsyncServer::removeConnection(HealthcheckHttpConnection* conn) {
+  delete conn;
```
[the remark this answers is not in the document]
Shikugawa: https://github.com/istio-ecosystem/authservice/pull/174#discussion_r732622520

PR #179, opened 2021-10-19
Update bookinfo example with operator.yaml example for easier setup.

incfly:
This also fixes the indentation problem in the sample config...


incfly:
@Shikugawa FYI same as the other one. I have tested this manually myself. Let me know if you see any issues.

Issue #192, opened 2021-11-22
Make it possible to build FIPS authservice image.


incfly:
Authservice uses boringssl for several purpose. For a FIPS compliant consumer, this needs to be compiled with FIPS compliant way as well. This is to track all the necessary work.

incfly:
Now the build process is ready for FIPS. I am not so familiar with the release process. I took a quick glance but didn't see much info about the CI job or doc on releases.

@Shikugawa @ZackButcher Any pointers on how the release is managed and created? e.g. how is 0.4.1 created? Or  it's just: local dev machine build & push & release creation on Github?

incfly:
retitled, and release work will be tracked in https://github.com/istio-ecosystem/authservice/issues/197

Issue #155, opened 2021-08-23
introduce clang CI and fix current link error


Shikugawa:
After running `make build` with clang-10, we will see this link failure with messages as follows.

```
ERROR: /home/shikugawa/dev/authservice-test/src/main/BUILD:5:10: Linking of rule '//src/main:auth_server' failed (Exit 1): clang failed: error executing command /opt/llvm/bin/clang @bazel-out/k8-fastbuild/bin/src/main/auth_server-2.params

Use --sandbox_debug to see verbose messages from the sandbox clang failed: error executing command /opt/llvm/bin/clang @bazel-out/k8-fastbuild/bin/src/main/auth_server-2.params

Use --sandbox_debug to see verbose messages from the sandbox
bazel-out/k8-fastbuild/bin/external/boost/libboost_thread.a(thread.o):thread.cpp:function void boost::throw_exception<boost::bad_weak_ptr>(boost::bad_weak_ptr const&) [clone .isra.0]: error: relocation refers to global symbol "construction vtable for boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::bad_weak_ptr> >-in-boost::wrapexcept<boost::bad_weak_ptr>", which is defined in a discarded section
  section group signature: "_ZTVN5boost10wrapexceptINS_12bad_weak_ptrEEE"
  prevailing definition is from bazel-out/k8-fastbuild/bin/src/common/http/_objs/http/http.pic.o
clang-10: error: linker command failed with exit code 1 (use -v to see invocation)
Target //src/main:auth_server failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 88.731s, Critical Path: 46.83s
INFO: 2194 processes: 9 internal, 2185 linux-sandbox.
FAILED: Build did NOT complete successfully
```

Fix this problem and introduce clang CI as current gcc.

incfly:
Resolved as the `--config=clang` becomes an option.

PR #193, opened 2021-11-22
Add clang.bazelrc setup from envoy.

incfly:

#192 

FIPS boringssl requires clang to compile. This is the first step to setup the clang copmilation option.

### DIFF
--- a/bazel/boost.bzl
+++ b/bazel/boost.bzl
@@ -13,6 +13,10 @@ def boost_library(name, deps = []):
                     "libboost_{}.a".format(name),
                     "libboost_{}.dylib".format(name)
                 ],
+                "clang": [
+                    "libboost_{}.a".format(name),
+                    "libboost_{}.so.{}".format(name, BOOST_VERSION),
+                ],
                 "//conditions:default": [
                     "libboost_{}.a".format(name),
                     "libboost_{}.so.{}".format(name, BOOST_VERSION),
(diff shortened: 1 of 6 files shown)


bazel/setup_clang.sh
```diff
@@ -0,0 +1,34 @@
...
+
+echo "# Generated file, do not edit. If you want to disable clang, just delete this file.
+build:clang --action_env='PATH=${PATH}'
+build:clang --action_env=CC=clang
+build:clang --action_env=CXX=clang++
+build:clang --action_env='LLVM_CONFIG=${LLVM_PREFIX}/bin/llvm-config'
+build:clang --repo_env='LLVM_CONFIG=${LLVM_PREFIX}/bin/llvm-config'
+build:clang --linkopt='-L$(llvm-config --libdir)'
+build:clang --linkopt='-Wl,-rpath,$(llvm-config --libdir)'
+build:clang --define clang_cl=1
```
incfly, line 26: This is the place that I diverts from the envoy upstream. the `clang_cl` is used in the boost config_settings option. Feel there could be a better way. @lizan

lizan: clang_cl is the Windows option, do we use that?

incfly: We don't. I only need a way to be able to defer that it's `--config=clang`. figured it out. Changed the BUILD.boost config_settings to use `value: {'action_env': 'CXX=clang++'}`

bazel/boost.bzl
```diff
@@ -13,6 +13,10 @@ def boost_library(name, deps = []):
                     "libboost_{}.a".format(name),
                     "libboost_{}.dylib".format(name)
                 ],
+                "clang": [
```
Shikugawa, line 16: Do we need it here? This seems the same as `//conditions:default`.

incfly: there's a difference, it adds `toolset=clang`. without this, boost compilation fails, https://github.com/istio-ecosystem/authservice/issues/155

Shikugawa (approved):

Issue #197, opened 2021-11-28
Create authservice 0.5.0 release


incfly:
We have made several changes recently. Including

- Fips build
- https://github.com/istio-ecosystem/authservice/pull/187
- dynamic jwks support.
- health check for jwks

This is worth a new release. The release should

- [x] Create release branch, release-0.5
- [x] Build the new image, with fips and non fips options. The https://github.com/istio-ecosystem/authservice/pull/196 is in flight.
- [x] Test both are working end to end with the bookinfo example, we provided. https://github.com/istio-ecosystem/authservice/tree/master/bookinfo-example
- [x] Document the release notes. 
- [x] Release via Github release, a previous one e.g. https://github.com/istio-ecosystem/authservice/releases/tag/0.4.1

Shikugawa:
Finally, we released v0.5.0. https://github.com/istio-ecosystem/authservice/releases/tag/0.5.0

PR #199, opened 2021-12-01
Enable CI for clang with FIPS mode.

Shikugawa:
https://github.com/istio-ecosystem/authservice/issues/197

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -14,11 +14,8 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Setup clang-format
-      env:
-        CLANG_VERSION: '12.0.0'
       run: |
-        wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${CLANG_VERSION}/clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-20.04.tar.xz
-        mkdir -p /opt/llvm && tar -xvf clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-20.04.tar.xz -C /opt/llvm --strip-components 1
+        sudo sh -c ${PWD}/bazel/install-clang-ci.sh
         git clone https://github.com/Sarcasm/run-clang-format.git
     - name: Run clang-format
       run: find ./ -iname "*.h" -o -iname "*.cc" | xargs ./run-clang-format/run-clang-format.py --clang-format-executable=/opt/llvm/bin/clang-format
@@ -46,8 +43,7 @@ jobs:
       env:
         CLANG_VERSION: '12.0.0'
       run: |
-        wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${CLANG_VERSION}/clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-20.04.tar.xz
-        mkdir -p /opt/llvm && tar -xvf clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-20.04.tar.xz -C /opt/llvm --strip-components 1
+        sudo sh -c ${PWD}/bazel/install-clang-ci.sh
         bazel/setup_clang.sh /opt/llvm
     - name: make test
       run: |
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -14,15 +14,15 @@ jobs:
     steps:
     - uses: actions/checkout@v1
     - name: Setup clang-format
+      env:
+        CLANG_VERSION: '12.0.0'
       run: |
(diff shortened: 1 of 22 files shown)


bazel/bazel.bzl
```diff
@@ -1,9 +1,20 @@
...
 
-def xx_binary(name, deps = [], srcs = [], copts = [], defines = []):
-    native.cc_binary(name = name, deps = deps, srcs = srcs, copts = _DEFAULT_COPTS + copts, defines = defines)
+def authsvc_cc_test(name, deps = [], srcs = [], data = []):
+    cc_test(
+        name = name,
+        deps = deps,
+        srcs = srcs,
+        data = data,
+        linkstatic = True
```
Shikugawa, line 19: I ran into a problem due to the dynamic library symbol resolution. This is a workaround strategy to resolve that. It doesn't have any effect.

incfly: So dyanmic lib resolution would fail for the test to depend on fips boringssl? Do you mind pasting the error message?
Can you explain why `linkstatic=true` is a workaround? In another words, do we need better solution later?

test/common/session/BUILD
```diff
@@ -1,3 +1,6 @@
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//bazel:bazel.bzl", "authsvc_cc_test")
+
 cc_library(
```
incfly, line 4: does this have to be `cc_library`? Can this be `authsvc_cc_library` as the rest?

.github/workflows/workflow.yml
```diff
@@ -14,15 +14,15 @@ jobs:
...
     - uses: actions/checkout@v1
     - name: Setup clang-format
+      env:
+        CLANG_VERSION: '12.0.0'
       run: |
-        wget https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.1/clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-        tar -xvf clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04.tar.xz
-        sudo mv ./clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04/bin/clang-format /usr/local/bin
-        rm -rf clang+llvm-10.0.1-x86_64-linux-gnu-ubuntu-16.04/
+        wget https://github.com/llvm/llvm-project/releases/download/llvmorg-${CLANG_VERSION}/clang+llvm-${CLANG_VERSION}-x86_64-linux-gnu-ubuntu-20.04.tar.xz
```
incfly, line 20: can we extract this into a script like `bazel/install-clang-ci.sh`? 

This way, it's easier to reproduce the same environment locally (if someone need to debug the CI); and
we don't have to duplicate the `wget` commands in line 20 and line 49.

test/service/BUILD
```diff
@@ -21,5 +22,4 @@ cc_test(
         "//test/filters/oidc:mocks",
         "@com_google_googletest//:gtest_main",
     ],
-    linkstatic = select({"@boost//:osx": True, "//conditions:default": False}), # workaround for not being able to figure out how to link dynamically on MacOS
```
incfly, line 24: does this mean we will not be able to run the test in Mac OSX? which is fine to me, just want to ocnfirm.

incfly:
thanks. LGTM overall. Already see the CI have the new flow running. Just a few comments.

test/service/BUILD
Shikugawa: We should `linkstatic=True` on macOS. This change includes that to be True anytime. So we don't have to be concern.

bazel/bazel.bzl
Shikugawa: Yes. The logs posted here. https://gist.github.com/Shikugawa/0ff7ef056cf6fdb2605ad81fcb0be814
I expect more clear linkopt for dynamic resolution. But I couldn't find a clear way of it. And we don't have to stick to be dynamic link here. This is why I changed to use `linkstatic=True`.

incfly: Got it. Could you make a comment above `linkstatic=true` with the reason why it's used? something like `We choose to use static link because boringssl FIPS build seem not be able to resolved for unit test, <gist-link> (optional)`.

.github/workflows/workflow.yml
```diff
@@ -34,3 +31,21 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Setup Bazel
+      run: sudo build/install-bazel.sh
+    - name: Install dependencies
+      run: sudo apt-get update && sudo apt-get -y install make cmake ninja-build build-essential
+    - name: Setup clang
+      env:
+        CLANG_VERSION: '12.0.0'
```
incfly, line 44: we can omit this env var right? `install-clang-ci.sh` currently hardcoded this as 12.0.0 as well.

incfly (commented):
Just one last comment.

Issue #180, opened 2021-10-21
refactoring HTTP client


Shikugawa:
Previous HTTP client code has some problems.

1. It doesn't have transport socket options. So we can't switch raw or tls socket to use in requesting.
2. Many lines of code are included in the same function. We should separate them.

And also, It has some points to cleanup by @incfly 

- https://github.com/istio-ecosystem/authservice/pull/174#discussion_r733036936
- https://github.com/istio-ecosystem/authservice/pull/174#discussion_r733038206
- https://github.com/istio-ecosystem/authservice/pull/174#discussion_r733043346

PR #201, opened 2021-12-15
wip

Shikugawa:


Shikugawa:
We need to do #180 to add option to skip cert verify to use self-signed certs in e2e.

PR #202, opened 2022-01-07
add option to disable remote ca verify

Shikugawa:
This is needed for #201.

### DIFF
--- a/src/common/http/http.h
+++ b/src/common/http/http.h
@@ -48,7 +48,7 @@ class PathQueryFragment {
 };
 
 struct TransportSocketOptions {
-  // the ca cert to be trusted in the http call
+  // the ca cert to be trusted in the http call, in PEM format.
   std::string ca_cert_;
   // whether or not to validate peer cert.
   bool verify_peer_ = true;
--- a/src/common/http/http.h
+++ b/src/common/http/http.h
@@ -12,6 +12,8 @@
 
 #include "absl/strings/string_view.h"
 #include "absl/types/optional.h"
+#include "boost/beast/core/tcp_stream.hpp"
+#include "boost/beast/ssl/ssl_stream.hpp"
 
 namespace beast = boost::beast;  // from <boost/beast.hpp>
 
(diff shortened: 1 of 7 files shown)


test/filters/oidc/oidc_filter_test.cc
```diff
@@ -1128,7 +1138,9 @@ google::rpc::Code OidcFilterTest::MakeRequestWhichWillCauseTokenRetrieval(
   auto raw_http = common::http::response_t(
       new beast::http::response<beast::http::string_body>());
   raw_http->result(beast::http::status::ok);
-  EXPECT_CALL(*mocked_http, Post(Eq(token_uri), _, _, Eq("some-ca"),
+  const common::http::TransportSocketOptions opt{.ca_cert_ = "some-ca",
```
incfly, line 1141: will it make sense to have a `kDefaultTransportSocketOpts` constant defined in the beginning?

src/common/http/http.h
```diff
@@ -45,6 +47,17 @@ class PathQueryFragment {
   inline bool HasFragment() const { return !fragment_.empty(); }
 };
 
+struct TransportSocketOptions {
+  // the ca cert to be trusted in the http call
```
incfly, line 51:
```suggest
  // the ca cert to be trusted in the http call, in PEM format.
```

incfly: If I understand correctly, it's inlined PEM format.

incfly (commented):
LGTM, just a few nitpicks.

bruegth:
@incfly 
Did you forgot this RFC thing (https://github.com/istio-ecosystem/authservice/pull/126#issuecomment-934655392) ?

But nice to have this now. 👍🏻

Shikugawa:
> @incfly Did you forgot this RFC thing ([#126 (comment)](https://github.com/istio-ecosystem/authservice/pull/126#issuecomment-934655392)) ?
> 
> But nice to have this now. 👍🏻

We have made it clear to the API that this feature should only be used in a test environment. (This is required for integration testing with self-signed certificates.)

bruegth:
@Shikugawa Sure, but result would be the same or? 

And if an ODICP instance is behind an Reverse Proxy (TLS endpoint) on same host, it still makes no sense to put an certificate onto it.

Shikugawa:
@bruegth The IDP setting of authservice is implemented to accept only HTTPS, and adding an option to allow HTTP to it is not desirable. Therefore, when configuring an external IDP for integration testing, a TLS certificate should always be set. However, it is practical to use self-signed certificates in a test environment, and for that reason, we should be able to disable certificate validation.

incfly:
When reviewing PR, I actually did worry about this can cause misunderstanding like this, since it's defined as part of the API proto. @Shikugawa for integration test we could provide a way to specify additional root (even if it's self signed). Istio does the similiar things for its JWKS endpoint HTTPS fetching.

Shikugawa:
I also had concerns about providing this feature as an API, so I considered several other ways (e.g., providing it as a runtime option, or switching it in a compile option), but I thought that an API would be fine as long as it is clearly documented. However, if @incfly has other concerns, we can provide a better way. We can just not include this change in the next release.

PR #204, opened 2022-01-21
introduce integration test for authorization code grant flow

Shikugawa:


incfly:
Thanks! I have a quick comment: people working or need to try out this project are mostly from Istio/k8s background. So they are familiar with k8s, and we use k8s cluster in bookinfo example as well.

Could we use kind cluster as testing environment? So that devs above don't have to learn another tool (docker compose) to debug e2e test failure in future.

Moreover, let's say later we want to include helm release package, we want to use the same config in our e2e tests to guard our release. These configs will be k8s config.

Shikugawa:
@ZackButcher @incfly I found that `GITHUB_TOKEN` has only permission to read `ghcr.io` on the organization level. Thus we can't run to push built images to the registry on gha. But I don't belong to this organization. (only having repository level admin permission) Therefore, I ask the members in this organization to change the `GITHUB_TOKEN` default permission to allow both write and read.

incfly:
I checked some doc and our settings, seem only authservice repo itself need to have the read/write permission, which is already the case?

https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#upgrading-a-workflow-that-accesses-ghcrio

perm

nacx:
This is already covered in the latest versions of authservice

Issue #205, opened 2022-01-21
"Retrieve token: invalid form" in a loged-in session window.


incfly:
Both @nacx and @vikaschoudhary16 mentioned they run into issues that:

1. callback URL redirection fails with empty page, and 403.
2. the authservice prints out the error log such as `retrieve token: invalid form`.
3. after using a incognito window, it works.

After some instrumentation, @vikaschoudhary16 found that could be due to a encoding formation issue.

---

what happens is that if I am doing fresh login into google acount as part of getting authz code, then in the callback api, broswer provides code in the encoded format, which is expected by authservice and it works

if I am already logged in and I retry getting authorization code for a new session, this time browser provides authz code, in the query params of callback uri, in the decoded format which has charatcer like /

this fails in authservice as authservice expects only encoded format

like `%2f` for /

will it be okay if we, authservice,  look for reserved chaacters in the auth code and if found.. that means code is already in decoded form and skip decoding

incfly:
Given this has appeared in two recent setup, I think it's worth looking a bit further.

incfly:
@nacx  @vikaschoudhary16 You could try this image gcr.io/jianfeih-test-321318/authservice:0124a to replace with your authservice deployment. This should fix the problem.

nacx:
I'll try this week. Thanks!
JFTR, I switched to Auth0 for my demo and the issue does not manifest there.

vikaschoudhary16:
> @nacx @vikaschoudhary16 You could try this image gcr.io/jianfeih-test-321318/authservice:0124a to replace with your authservice deployment. This should fix the problem.

 It should work Jianfei. I trid a hack similar to the changes in the PR couple days back and that got it working.
![image](https://user-images.githubusercontent.com/11936386/151101706-3ad50108-4f9c-4a47-b24d-df6fd6cd79e7.png)

PR #206, opened 2022-01-24
Fix url encoding handling for code parameter.

incfly:

Fix https://github.com/istio-ecosystem/authservice/issues/205

### DIFF
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -66,6 +66,8 @@ bool IsFormDataSafeCharacter(const char character) {
 // https://developers.google.com/identity/protocols/oauth2/openid-connect#sendauthrequest
 // [2] https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.11.
 // [3] https://www.rfc-editor.org/rfc/rfc3986#page-12
+// TODO(incfly): move to a separate library specific for OIDC instead of putting
+// in http lib.
 bool IsOIDCCodeSafeCharacter(const char character) {
   return IsUrlSafeCharacter(character) || (character == '/');
 }
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -53,6 +53,23 @@ bool IsFormDataSafeCharacter(const char character) {
   return IsUrlSafeCharacter(character) || (character == '+');
 }
 
+// We found that at least accounts.google.com[1] could contain `/` in the
+// authorization code, as part of OAuth callback URl query param.
+// such as `..../?code=4/Abc38.
+// RFC 6749[2] specifies "code" could be `VSCHAR`, containing all printable
+// ascii chars. RFC 3986[3] about URI specifies that:
+// If a reserved character is found in a URI component and
+//  no delimiting role is known for that character, then it must be
+//  interpreted as representing the data octet corresponding to that
+//  character's encoding in US-ASCII.
+// [1]
+// https://developers.google.com/identity/protocols/oauth2/openid-connect#sendauthrequest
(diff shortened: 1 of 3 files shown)


src/common/http/http.cc
```diff
@@ -53,6 +53,23 @@ bool IsFormDataSafeCharacter(const char character) {
...
+// ascii chars. RFC 3986[3] about URI specifies that:
+// If a reserved character is found in a URI component and
+//  no delimiting role is known for that character, then it must be
+//  interpreted as representing the data octet corresponding to that
+//  character's encoding in US-ASCII.
+// [1]
+// https://developers.google.com/identity/protocols/oauth2/openid-connect#sendauthrequest
+// [2] https://datatracker.ietf.org/doc/html/rfc6749#appendix-A.11.
+// [3] https://www.rfc-editor.org/rfc/rfc3986#page-12
+bool IsOIDCCodeSafeCharacter(const char character) {
```
Shikugawa, line 69: I think that `http.cc` should be independent from filter implementation. It seems this rule is OIDC specific. Thus we should move this to `oidc_filter.cc` or new file such as `oidc_query_decoder.cc` and pass it to `Http::DecodeQueryData`. For example, I prefer to implement that as follows:

```
class UrlQueryDecoder {
  virtual absl::string_view decode(absl::string_view) = 0;
};
```

```
class OidcUrlQueryDecoder : public UrlQueryDecoder {
    // implement...
};
```

But, we should prioritize to fix this problem, Therefore we should postpone this refactor as TODO.

vikaschoudhary16:
Accommodating just one character, `/`,  will not help for very long I think. As we discussed offline, in my case it was the client(browser) which is passing code in the decoded form. my authz code was having `%2f` in encoded form which in decoded form is `/`. Is not it possible that there is some other combination of letters in encoded form which when decode is some other key than the `/`. 
Instead of handling `/` individually, can we determine if the code is encoded or decode by checking presence of reserved keys and if already decoded, skip decoding.

src/common/http/http.cc
incfly: Add a TODO.

I think putting out of http lib is fine (this is not filter lib, but http, btw). but I'm not sure if we have to setup class/inheritance, especially if there's only one method, without actual state to maintain.
Pure util function seems work well and clear.

incfly:
> Accommodating just one character, /, will not help for very long I think. 

There's actually only a few char to be considred. `/`, `:` `=` etc. The `=` would not be considered valid. Also in other identity provider, the issue does not occur, say Auth0. I think it's fine to be conservative and keep track of what characters we allow explicitly. i.e. only add chars when we encounter one.

https://stackoverflow.com/questions/14856241/url-encoding-of-the-state-parameter-for-google-oauth2-gets-decoded-during-redire

src/common/http/http.cc
incfly: ping, is that sufficient for now? @Shikugawa

vikaschoudhary16:
> > Accommodating just one character, /, will not help for very long I think.
> 
> There's actually only a few char to be considred. `/`, `:` `=` etc. The `=` would not be considered valid. Also in other identity provider, the issue does not occur, say Auth0. I think it's fine to be conservative and keep track of what characters we allow explicitly. i.e. only add chars when we encounter one.
> 
> https://stackoverflow.com/questions/14856241/url-encoding-of-the-state-parameter-for-google-oauth2-gets-decoded-during-redire

sounds good!

src/common/http/http.cc
Shikugawa: Yes. SGTM.

Shikugawa (approved):

PR #208, opened 2022-02-04
Make sure we can build on macOS

dio:
This patch makes sure we can build on macOS (while limited for amd64 for now).

Several changes in this patch:

* Rename `BUILD` to `BUILD.bazel` due to case-insensitivity of (the default) macOS filesystem.
* Use `github.com/nelhage/rules_boost` to manage boost dependencies.
* Always run `bazel` through `bazelisk`.
* Set `.bazelversion` to `4.2.2` (still can't update to `5.0.0` yet).
* Format and lint `bazel` files with `buildifier`. (Sorry for the formatting noise! I need that since I was trying to fix #176), and wanting to be sure it is lint-ed.
* Add a macOS-specific workflow for now (will merge with the main workflow later).
* Fixes https://github.com/istio-ecosystem/authservice/issues/176.

Notes to the reviewer: I can try to clean up the Makefile more and streamline the management of `clang` (including `clang-format`), and other required toolings: if that makes sense of sending that kind of PR(s).

Signed-off-by: Dhi Aurrahman <dio@rockybars.com>

PR #210, opened 2022-02-08
Add release workflow

dio:
This adds release workflow to upload the built binaries (as tar.gz archives) triggered by tag creation. The tag that triggers this workflow should match: `v[0-9]+.[0-9]+.[0-9]+**` For example: `v0.5.0 v0.5.1-rc2`. A sample of release in the forked repo: https://github.com/3128px/authservice/releases/tag/v0.5.1-rc0, the triggered workflow run https://github.com/3128px/authservice/actions/runs/1810249439.

There is no attached release notes (not yet templated), but I think we can do a follow-up on this, or do editing for each published release.

As bonuses, 

1. `make help` with colors!

image

2. Fixed the `make format` to run against all eligible files.
3. Cache `bazel` build files on CI.
4. The workflow is now simpler, spawning all possible combinations of `os` and `mode`. A sample of running it: https://github.com/3128px/authservice/actions/runs/1810175327 with `check` job as a pre-requisite.


dio:
Not sure why the DCO is failing. Let me take a look.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,104 @@
+name: "release"
```
incfly, line 1: We release docker image to ghcr registry as part of release. Shall that be done in this workflow?
not necessary to be done in this PR, just want to understand desired steps.

incfly (commented):
Thanks! Can we have how to kick off release process documented somwhere?

I checked but we don't seem have a suitable dev doc place. Maybe add a `DEVELOPMENT.md`?

Makefile
```diff
@@ -56,29 +66,55 @@ docker-from-scratch:
 docker-compile-env:
 	docker build -f build/Dockerfile.interactive-compile-environment -t authservice-build-env:$(USER) .
 
-bazel-bin/src/main/auth_server:
-    # Note: add --compilation_mode=dbg to the end of the next line to build a debug executable with `make docker-from-scratch`
+$(current_binary):
```
incfly, line 69: shall the line below be replaced with `(call bazel-build)`?

Makefile
```diff
@@ -56,29 +66,55 @@ docker-from-scratch:
...
+# Note: add --compilation_mode=dbg to the end of the next line to build a debug executable with `make docker-from-scratch`
```
incfly, line 70: I tried `make docker` after this PR seems got some error. maybe the docker rule needs update as well? (`fly` is my linux user name)

```
mkdir -p /home/fly/workspace/github.com/istio-ecosystem/authservice/.cache/tools/bazelisk /home/fly/workspace/github.com/istio-ecosystem/authservice/.cache/bazel
GOARCH=amd64 /usr/local/go/bin/go run github.com/bazelbuild/bazelisk@v1.11.0 --output_user_root=/home/fly/workspace/github.com/istio-ecosystem/authservice/.cache/bazel build   //src/main:auth_server
make: *** No rule to make target 'fly', needed by 'docker'.  Stop.
```

Makefile
```diff
@@ -56,29 +66,55 @@ docker-from-scratch:
...
 clang-format := $(prepackaged_tools_dir)/bin/clang-format
 
-build:
-	$(bazel) build $(BAZEL_FLAGS) //src/...
-
-run:
+build: ## Build the main binary
+	$(call bazel-build)
+
+# This should be overridden by current tag (with stripped "v") when running `make dist` on CI.
```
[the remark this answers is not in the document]
dio: Ah, yeah. We need to have something like: v0.5.1-rc0 and the one that will be used as part of the release archives file name is 0.5.1-rc0.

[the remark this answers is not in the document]
dio: Will update it.

bazel/BUILD.boost
```diff
@@ -1,80 +0,0 @@
-config_setting(
```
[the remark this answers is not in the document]
dio: We don't need this, since we replace it in the previous merged PR: https://github.com/istio-ecosystem/authservice/pull/208. We use github.com/nelhage/rules_boost to manage boost dependencies now.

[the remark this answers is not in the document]
dio: I think this is a leftover :).

dio:
> Thanks! Can we have how to kick off release process documented somwhere?

will do.

Makefile
dio: Got it will update.

.github/workflows/release.yaml
dio: Yup. Will do docker-related stuff after this PR, but I can take a look.

dio:
After some thinking, I will deal with build-container, docker-compose later after this PR. Since both of them requires more time for local testing.

Makefile
dio: I changed this to `make image`. I will deal with everything container in a follow up PR if you don't mind.

Makefile
dio: Updated.

Makefile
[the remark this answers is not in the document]
dio: Updated.

Makefile
incfly: np.

incfly:
tried with `make image` and it works. LGTM.

incfly (approved):

Issue #211, opened 2022-03-25
integration with ADFS fails due to unable to shutdown connection for ssl stream


incfly:
Documented an issue we are trying to fix.

Scenario: An ADFS server act as IdP; auth service configured with client id/secret as usual.
End user observed behavior: redirection is fine, but authservice can't reach out to idp to get id token, therefore HTTP request fails

```
In the authservice.log, looking at the following snippet shows the initial call to RetrieveToken,
connecting to ADFS, and then 2 minutes of waiting until RetrieveToken returns:

[2022-03-22 16:26:38.188] [console] [trace] RetrieveToken
[2022-03-22 16:26:38.188] [console] [trace] Post
[2022-03-22 16:26:38.528] [console] [info] Post: Trusting the provided certificate authority
[2022-03-22 16:26:38.530] [console] [info] Post: opening connection to https://adfs.example.com:443
[2022-03-22 16:27:26.823] [console] [info] operator(): Starting periodic cleanup (period of 60 seconds)
[2022-03-22 16:27:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain kafdrop
[2022-03-22 16:27:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain nifi
[2022-03-22 16:27:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain nifi-registry
[2022-03-22 16:28:26.823] [console] [info] operator(): Starting periodic cleanup (period of 60 seconds)
[2022-03-22 16:28:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain kafdrop
[2022-03-22 16:28:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain nifi
[2022-03-22 16:28:26.823] [console] [info] DoPeriodicCleanup: removing expired sessions from chain nifi-registry
[2022-03-22 16:28:42.787] [console] [info] RetrieveToken: Saving token response to session store

Looking at the Istio sidecar logs within the same timeframe (begin: 2022-03-22T16:26:38.620989Z) shows the following:
2022-03-22T16:26:38.620989Z debug envoy filter original_dst: New connection accepted
2022-03-22T16:26:38.621109Z debug envoy filter [C442] new tcp proxy session
2022-03-22T16:26:38.621128Z trace envoy connection [C442] readDisable: disable=true disable_count=0 state=0 buffer_length=0
[...]

PR #212, opened 2022-03-25
Change to shutdown the connection directly in case the server do not participate close.

incfly:
Fixes https://github.com/istio-ecosystem/authservice/issues/211

After this PR, the issue is resolved.

### DIFF
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -503,23 +503,13 @@ response_t HttpImpl::Post(
     response_t res(new beast::http::response<beast::http::string_body>);
     beast::http::async_read(stream, buffer, *res, yield);
 
-    // Gracefully close the socket.
-    // Receive an error code instead of throwing an exception if this fails, so
-    // we can ignore some expected not_connected errors.
-    boost::system::error_code ec;
-    stream.async_shutdown(yield[ec]);
-
-    if (ec) {
-      // not_connected happens sometimes so don't bother reporting it.
-      // stream_truncated also happens sometime and we choose to ignore the
-      // stream_truncated error, as recommended by the github thread:
-      // https://github.com/boostorg/beast/issues/824
-      if (ec != beast::errc::not_connected &&
-          ec != boost::asio::ssl::error::stream_truncated) {
-        spdlog::error("{}: HTTP error encountered: {}", __func__, ec.message());
-        return response_t();
-      }
-    }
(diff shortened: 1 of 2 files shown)


incfly:
@dio in case @Shikugawa is not avaialable.

src/common/http/http.cc
```diff
@@ -503,23 +503,13 @@ response_t HttpImpl::Post(
...
-        return response_t();
-      }
-    }
+    spdlog::trace("{}: closing connection, response payload size {}", __func__,
+                  res->payload_size().value());
+    // Close the socket. We already got the response we need so close the socket
+    // directly. We choose not to use `ssl_stream::async_shutdown` since in some
+    // case, the HTTPS server does not participate with closing
+    // stream/connection. That would make the async_shutdown waiting forever.
+    beast::get_lowest_layer(stream).close();
```
Shikugawa, line 512: I got the cause of this change. This problem is rooted from asio's implementation and I think it is better to use timeout to wait ack of `close_notify` message. (But `async_shutdown` doesn't have this feature. ref https://github.com/chriskohlhoff/asio/issues/650)
My proposal is to note that this fix is like a workaround here and may be fixed in future work.

Shikugawa, line 512: Couldn't we handle error code here?

dio: I like the timeout idea.

incfly: I saw that post as well. And my initial attempt, https://github.com/istio-ecosystem/authservice/pull/212/commits/eeb57760a099a06f98a0fb1e51472029e3f855da was trying to use a timer, but I didn't managed to get it work properly, lack of C++ and boost familiarity ... :( 

Added a tracker and a TODO. https://github.com/istio-ecosystem/authservice/issues/214

incfly: close does not return error code AFAIK, https://www.boost.org/doc/libs/1_72_0/libs/beast/doc/html/beast/ref/boost__beast__basic_stream/close.html

Shikugawa: I think that https://github.com/istio-ecosystem/authservice/commit/eeb57760a099a06f98a0fb1e51472029e3f855da won't work because the timer callback won't be executed after `async_shutdown` will have done ;) This is why we should wait the fix of boost/asio internal. Thanks.

Shikugawa (approved):

PR #213, opened 2022-03-27
Fix make compose

dio:
This makes sure we can do `make compose`, and optionally build the binary (also the image) of `auth_server` (by feeding the `--build` flag: `COMPOSE_ARGS=--build make compose`).


src/service/async_service_impl.cc
```diff
@@ -15,7 +15,7 @@ namespace authservice {
 namespace service {
 
 namespace {
-constexpr uint16_t kHealthCheckServerPort = 10004;
+constexpr uint16_t kHealthCheckServerPort = 9091;
```
dio, line 18: I'm not sure regarding this, seems like we had the same health check port (with the main one? I kept getting the port-in-use error).

dio:
I think once:

- #215

is merged, I'll revise this a bit.

PR #215, opened 2022-03-30
Append -static-libgcc on Linux

dio:
This patch makes sure we do: "-static-libgcc" on Linux (for the stripped binary). We also use `--config libc++` (instead of `--config clang`) to link against `libc++` on CI (Linux).

This also adds a workflow to publish the Docker image. 

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -65,12 +65,12 @@ jobs:
         # This downloads the required clang tooling when it is not downloaded yet.
         run: |
           make clang.bazelrc
-          echo "BAZEL_FLAGS=--config=clang" >> $GITHUB_ENV
+          echo "BAZEL_FLAGS=--config=libc++" >> $GITHUB_ENV
 
       # Set BAZEL_FLAGS to FIPS mode only when it is required.
       - name: Setup FIPS mode
         if: matrix.mode == 'clang-fips'
-        run: echo "BAZEL_FLAGS=--config=clang --define=boringssl=fips" >> $GITHUB_ENV
+        run: echo "BAZEL_FLAGS=--config=libc++ --define=boringssl=fips" >> $GITHUB_ENV
 
       - name: Create artifacts # We strip the "v"-prefix from the current tag.
         run: VERSION=${GITHUB_REF#refs/tags/v} MODE=${{ matrix.mode }} make dist
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -22,6 +22,7 @@ jobs:
           - "ubuntu-20.04"
         mode:
           - "default"
+          # By default we use libc++.
           - "clang"
           - "clang-fips"
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -65,16 +66,34 @@ jobs:
...
+        run: |
+          make requirestatic
+
+      - name: Login to GitHub Container Registry
+        if: runner.os == 'Linux' && matrix.mode == 'clang'
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.PAT }}
```
dio, line 90: @incfly to make it work, we need your help to add this [PAT](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) to the secrets of this repo. Thanks!

test/exe/require_static.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Copied from https://github.com/envoyproxy/envoy/blob/a12869fa9e9add4301a700978d5489e6a0cc0526/test/exe/envoy_static_test.sh.
```
incfly, line 3: A few words in the beginning to explain the purpose of the script? something like `this scripts checks if the provided binary is static linked. Fails if not.`.

test/exe/require_static.sh
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ $(uname) == "Darwin" ]]; then
+  echo "macOS doesn't support statically linked binaries, skipping."
+  exit 0
+fi
+
+# We can't rely on the exit code alone, since ldd fails for statically linked binaries.
```
incfly, line 10: Yeah I noticed that as well when using `ldd` to check a go program. it instead complains the input file is not valid :)

test/exe/require_static.sh
incfly: never mind. i think this one is pretty self explained.

incfly (approved):
Thanks!

.github/workflows/release.yaml
incfly: oops. I forgot Approving the PR makes the PR merge automatically. let me send a fix to update the variable name used here.

bazel/bazel.bzl
```diff
@@ -2,13 +2,17 @@
...
 
+# envoy_stdlib_deps appends "-static-libgcc" on Linux.
+load("@envoy//bazel:envoy_internal.bzl", "envoy_stdlib_deps")
+
 _DEFAULT_COPTS = ["-Wall", "-Wextra"]
 
 def authsvc_cc_library(name, deps = [], srcs = [], hdrs = [], copts = [], defines = [], includes = [], textual_hdrs = [], visibility = None):
     cc_library(name = name, deps = deps, srcs = srcs, hdrs = hdrs, copts = _DEFAULT_COPTS + copts, defines = defines, includes = includes, textual_hdrs = textual_hdrs, visibility = visibility)
 
+# By default, we always do linkstatic: https://docs.bazel.build/versions/main/be/c-cpp.html#cc_binary.linkstatic.
```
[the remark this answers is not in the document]
dio: Ah yes, seems like to force linking libgcc statically you need to ask for -static-libgcc. I haven't tried to do `fully_static_link` though since it will do static linking for glibc as well which is not recommended.

PR #218, opened 2022-04-03
Build on ubi8 base image

dio:
Signed-off-by: Dhi Aurrahman <dio@rockybars.com>

PR #219, opened 2022-04-05
Check on GLIBC requirements

dio:
This makes sure we can run on lte GLIBC_2.28.

Supersedes https://github.com/istio-ecosystem/authservice/pull/218.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,7 +203,7 @@ endef
 
 # Install clang from https://github.com/llvm/llvm-project. We don't support win32 yet as this script
 # will fail.
-clang-os                          = $(if $(findstring $(goos),darwin),apple-darwin,linux-gnu-ubuntu-20.04)
+clang-os                          = $(if $(findstring $(goos),darwin),apple-darwin,linux-gnu-ubuntu-16.04)
 clang-download-archive-url-prefix = https://$(subst llvmorg/clang+llvm@,releases/download/llvmorg-,$($(notdir $1)@v))
 $(clang):
 	@mkdir -p $(dir $@)
--- a/Makefile
+++ b/Makefile
@@ -117,6 +117,7 @@ $(current_binary):
 ifeq ($(goos),linux)
 # Some -copt="-Wno-error=" is needed to suppress the error for gcc when compiling abseil and protobuf.
 gcc_w_no_error := --copt="-Wno-error=uninitialized" --copt="-Wno-error=deprecated-declarations" --copt="-Wno-error=maybe-uninitialized"
+clang_cxx      := --cxxopt="-D_GLIBCXX_USE_CXX11_ABI=1" --cxxopt="-DENVOY_IGNORE_GLIBCXX_USE_CXX11_ABI_ERROR=1"
 endif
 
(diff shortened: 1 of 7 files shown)


Tools.mk
```diff
@@ -1,5 +1,5 @@
 bazelisk@v     := github.com/bazelbuild/bazelisk@v1.11.0
-buildifier@v   := github.com/bazelbuild/buildtools/buildifier@4.2.5
+buildifier@v   := github.com/bazelbuild/buildtools/buildifier@5.0.1
```
dio, line 2: This is updated to bust the cache since currently, we don't cache per version. But we need to get a "new" clang.

test/exe/require_glibc.go
```diff
@@ -0,0 +1,39 @@
+package main
```
incfly, line 1: Just one comment above this to explain the purpose and invocation format? "This program checks the glibc version is higher than 2.27. Example `./require_glibc <binary>`."

test/exe/require_glibc.go
```diff
@@ -0,0 +1,39 @@
...
+
+func main() {
+	required, _ := version.NewVersion("2.27")
+	args := os.Args[1:]
+	cmd := exec.Command("objdump", "-T", args[0])
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
+	if err != nil {
+		log.Fatal("failed to check GLIBC version:", err)
```
incfly, line 22: Log line usually starts with capital letter. (line 31 as well)

Makefile
```diff
@@ -202,7 +202,7 @@ endef
 
 # Install clang from https://github.com/llvm/llvm-project. We don't support win32 yet as this script
 # will fail.
-clang-os                          = $(if $(findstring $(goos),darwin),apple-darwin,linux-gnu-ubuntu-20.04)
+clang-os                          = $(if $(findstring $(goos),darwin),apple-darwin,linux-gnu-ubuntu-16.04)
```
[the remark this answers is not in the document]
dio: Sorry, did you check the release page before commenting? https://github.com/llvm/llvm-project/releases/tag/llvmorg-12.0.0

test/exe/require_glibc.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+	scanner := bufio.NewScanner(strings.NewReader(out.String()))
+	for scanner.Scan() {
+		entry := scanner.Text()
+		if strings.Contains(entry, "GLIBC_") {
+			line := bufio.NewScanner(strings.NewReader(entry[strings.Index(entry, "GLIBC_")+len("GLIBC_"):]))
+			line.Split(bufio.ScanWords)
+			for line.Scan() {
+				v, _ := version.NewVersion(line.Text())
```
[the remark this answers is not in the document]
dio: I don't think it will fail, if it is it's a bug of objdump, but fine I'll add it.

test/exe/require_glibc.go
dio: I don't know about this "usual" thing. I'll update it.

incfly (approved):

PR #220, opened 2022-04-11
Skip X509_R_CERT_ALREADY_IN_HASH_TABLE error

dio:
Seems like this error can be ignored. References:

1. https://jira.mongodb.org/browse/TOOLS-1906?jql=Participants%20%3D%20kanant
2. https://github.com/facebook/folly/blob/d3354e2282303402e70d829d19bfecce051a5850/folly/ssl/OpenSSLCertUtils.cpp#L367-L368

### DIFF
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -424,7 +424,13 @@ response_t HttpImpl::Post(
           boost::asio::buffer(options.ca_cert_.data(), options.ca_cert_.size()),
           ca_ec);
       if (ca_ec) {
-        throw boost::system::system_error{ca_ec};
+        auto err = ERR_get_error();
+        // We can ignore this error. Reference:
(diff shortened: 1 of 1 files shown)


src/common/http/http.cc
```diff
@@ -424,7 +424,13 @@ response_t HttpImpl::Post(
           boost::asio::buffer(options.ca_cert_.data(), options.ca_cert_.size()),
           ca_ec);
       if (ca_ec) {
-        throw boost::system::system_error{ca_ec};
+        auto err = ERR_get_error();
```
incfly, line 427: how about still log it via trace log?

incfly (approved):
just found another case, https://forge.etsi.org/rep/cyber/103523_MSP/tlmsp/tlmsp-curl/-/commit/a0f3eaf25dd2c4acf1f49d28d5de4f1765521709

PR #221, opened 2022-04-11
Move the VERSION env to GITHUB_ENV.

incfly:

As suggested by @dio 
https://github.com/istio-ecosystem/authservice/commit/28404a8f3f9e11ff14e81d6650ffc72245355653#commitcomment-70845200


incfly:
@dio

incfly:
and yeah i noticed that the image tag we generate are actually suffix with some hexical string.. instead of 0.5.1

PR #223, opened 2022-07-26
Fix the unmatched request handling by set the grpc status in payload.

### LINKED ISSUE #140: Authservice should implicitly deny requests that don't match config prefixes
## Issue

Currently, a request to a workload that is authservice-enabled that does *not* match one of the prefixes in the authservice config will allow the request.

## Recreation

1. Spin up authservice with ext_authz envoyfilter
2. Deploy Istio gateway/virtualservice for application that does *not* match chain in authservice config
3. Deploy workload matching envoyfilter selector using above virtualservice
4. Access the application using the VS domain/routes, traffic will be allowed even though it isn't specified in config.

## Expected behavior

If a workload is authservice-enabled and it doesn't match a filter in the chain, it should be denied.

incfly:
Fix #140 

Previously we only return the grpc status, but not set the actual response payload, `response.status`.
Verified works manually e2e with `allow_unmatched_request: true/false` having different effect.

### DIFF
--- a/src/service/async_service_impl.h
+++ b/src/service/async_service_impl.h
@@ -12,6 +12,7 @@
 #include "common/config/version_converter.h"
 #include "config/config.pb.h"
 #include "envoy/common/exception.h"
+// #include "envoy/grpc/status.h"
 #include "envoy/service/auth/v2/external_auth.grpc.pb.h"
 #include "envoy/service/auth/v3/external_auth.grpc.pb.h"
 #include "src/common/http/http.h"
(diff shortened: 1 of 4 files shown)


incfly:
@dio @Shikugawa for review.

src/service/async_service_impl.h
```diff
@@ -12,6 +12,7 @@
 #include "common/config/version_converter.h"
 #include "config/config.pb.h"
 #include "envoy/common/exception.h"
+// #include "envoy/grpc/status.h"
```
lizan, line 15: remove?

Shikugawa (approved):

Shikugawa:
@incfly I observed a test failure in master... https://github.com/istio-ecosystem/authservice/runs/7541606968?check_suite_focus=true

incfly:
@Shikugawa suspect it's some build flakiness, rerunning, https://github.com/istio-ecosystem/authservice/runs/7545016277?check_suite_focus=true

PR #228, opened 2022-11-10
Use @envoy//bazel:boringssl

dio:
This makes sure we use @envoy//bazel:boringssl.

### DIFF
--- a/bazel/rules_boost.patch
+++ b/bazel/rules_boost.patch
@@ -1,0 +1,14 @@
+# To make sure we use the provided @envoy//bazel:boringssl.
+diff --git a/BUILD.boost b/BUILD.boost
+index 4354973..1c11525 100644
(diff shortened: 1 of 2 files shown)


bazel/rules_boost.patch
```diff
@@ -0,0 +1,14 @@
+# To make sure we use the provided @envoy//bazel:boringssl.
+diff --git a/BUILD.boost b/BUILD.boost
+index 4354973..1c11525 100644
+--- a/BUILD.boost
++++ b/BUILD.boost
+@@ -481,7 +481,7 @@ cc_library(
+     visibility = ["//visibility:public"],
+     deps = [
+         ":asio",
+-        "@openssl//:ssl",
```
dio, line 10: This is still boringssl, but a different version.

Shikugawa (approved):

PR #229, opened 2022-11-11
Ignore CERT_ALREADY_IN_HASH_TABLE error

sbko:
Check ec value for CERT_ALREADY_IN_HASH_TABLE instead of using ERR_get_error.

### DIFF
--- a/src/common/http/http.cc
+++ b/src/common/http/http.cc
@@ -424,11 +424,11 @@ response_t HttpImpl::Post(
           boost::asio::buffer(options.ca_cert_.data(), options.ca_cert_.size()),
           ca_ec);
       if (ca_ec) {
-        auto err = ERR_get_error();
+        std::string ca_message = ca_ec.message();
+        bool hash_error = ca_message.find("CERT_ALREADY_IN_HASH_TABLE") != std::string::npos;
         // We can ignore this error. Reference:
         // https://github.com/facebook/folly/blob/d3354e2282303402e70d829d19bfecce051a5850/folly/ssl/OpenSSLCertUtils.cpp#L367-L368.
-        if (ERR_GET_LIB(err) != ERR_LIB_X509 ||
-            ERR_GET_REASON(err) != X509_R_CERT_ALREADY_IN_HASH_TABLE) {
+        if (!hash_error) {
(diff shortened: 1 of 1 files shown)


src/common/http/http.cc
```diff
@@ -424,11 +424,11 @@ response_t HttpImpl::Post(
...
-        auto err = ERR_get_error();
+        std::string ca_message = ca_ec.message();
+        bool hash_error = ca_message.find("CERT_ALREADY_IN_HASH_TABLE") != std::string::npos;
         // We can ignore this error. Reference:
         // https://github.com/facebook/folly/blob/d3354e2282303402e70d829d19bfecce051a5850/folly/ssl/OpenSSLCertUtils.cpp#L367-L368.
-        if (ERR_GET_LIB(err) != ERR_LIB_X509 ||
-            ERR_GET_REASON(err) != X509_R_CERT_ALREADY_IN_HASH_TABLE) {
+        if (!hash_error) {
           throw boost::system::system_error{ca_ec};
         }
```
dio, line 433: Probably something like:

```cpp
        // X509_R_CERT_ALREADY_IN_HASH_TABLE can be ignored.
        // Reference:
        // https://github.com/facebook/folly/blob/d3354e2282303402e70d829d19bfecce051a5850/folly/ssl/OpenSSLCertUtils.cpp#L367-L368.
        if (ca_ec.category() != boost::asio::error::get_ssl_category() ||
            ERR_GET_REASON(ca_ec.value()) !=
                X509_R_CERT_ALREADY_IN_HASH_TABLE) {
          throw boost::system::system_error{ca_ec};
        }
```

I'm checking this on my local.

dio: So the asio's `ctx.add_certificate_authority` has already done the `ERR_get_error`, and wrapped it into an `error_code` "object".

dio: @sbko when you have time, could you help to test: https://github.com/3128px/authservice/releases/download/0.5.3-test2/auth_server_linux_amd64_clang-fips_0.5.3-test2.tar.gz? Thanks!

sbko: works!

sbko:
closed in favor of #230

PR #231, opened 2022-11-13
Enable FIPS for Linux only

dio:
The FIPS mode is only relevant for clang and Linux x68_64.


Shikugawa (approved):
LGTM

PR #235, opened 2023-05-03
Pass the trusted CA from the OIDC config to the HTTP GET call just like POST

asa-yeamans:
The GET call does not provide the trusted CA cert in `opt` to the HTTP GET call. Setting `ca_cert` to the provided value in the OIDC config

### DIFF
--- a/config/oidc/config.proto
+++ b/config/oidc/config.proto
@@ -93,6 +93,11 @@ message OIDCConfig {
         // true in any other cases.
         // Optional.
         bool skip_verify_peer_cert = 3;
+
+        // When specified, the Authservice will trust the specified Certificate Authority when performing HTTPS calls to
+        // the Token Endpoint of the OIDC Identity Provider.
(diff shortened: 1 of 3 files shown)


dio:
Could you fix DCO? I think this is a legit change. Thanks!

asa-yeamans:
Added DCO bits

jbraswell:
Any updates here? We'd love to have this fix.

asa-yeamans:
I am unable to build it locally due to issues creating a build environment, so I haven't been able to verify the changes build and subsequently that they perform as expected. The changes themselves are minimal so I presume they work, but assistance in verifying that would be helpful.

nacx:
This should have been addressed in v1.0.0 already, as part of the Go rewrite. Could you give it a try?

nacx:
This is already addressed in the latest versions of authservice, but please reopen if needed.

PR #247, opened 2024-04-12
Infer the JWS signing algorithm name by looking at the provided key

erik-h:
This handles cases where the JWS message or the key do not have a proper `alg` header. The `alg` header is [optional](https://datatracker.ietf.org/doc/html/rfc7517#section-4.4), so some identity providers may not supply it (such as [Microsoft Identity](https://login.microsoftonline.com/common/discovery/v2.0/keys)).

### DIFF
--- a/internal/authz/oidc.go
+++ b/internal/authz/oidc.go
@@ -621,7 +621,7 @@ func (o *oidcHandler) isValidIDToken(ctx context.Context, log telemetry.Logger, 
 		return false, codes.Internal
 	}
 
-	if _, err := jws.Verify([]byte(idTokenString), jws.WithKeySet(jwtSet)); err != nil {
+	if _, err := jws.Verify([]byte(idTokenString), jws.WithKeySet(jwtSet, jws.WithInferAlgorithmFromKey(true))); err != nil {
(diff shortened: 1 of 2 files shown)


internal/authz/oidc.go
```diff
@@ -621,7 +621,7 @@ func (o *oidcHandler) isValidIDToken(ctx context.Context, log telemetry.Logger,
 		return false, codes.Internal
 	}
 
-	if _, err := jws.Verify([]byte(idTokenString), jws.WithKeySet(jwtSet)); err != nil {
+	if _, err := jws.Verify([]byte(idTokenString), jws.WithKeySet(jwtSet, jws.WithInferAlgorithmFromKey(true))); err != nil {
```
nacx, line 624: Can you add a comment explaining why the infer algorithm is needed, so we don't accidentally remove it in the future?

nacx (commented):
Thank you! This LGTM!
Mind addressing the lint errors and adding some comments?

erik-h:
@nacx thanks for the review! I've added a couple comments and fixed the linter issues.

nacx (approved):

nacx:
Thx!

PR #249, opened 2024-04-16
Use the OIDC Discovery end session endpoint if present

nacx:
Currently, the end session URI present in the OIDC Discovery response was not used. This PR fixes this and uses the URI if none was configured.


sergicastro (approved):
Nice, good catch

PR #250, opened 2024-04-17
Add a logger to log the calls to the Identity Provider

nacx:
This adds the `idp` logger that will log all the calls made to the Identity Provider such as the token endpoint calls. It also adds more information to some of the existing loggers. Currently, there are no means to see what requests are being made as part of the OIDC flow, and troubleshooting the responses from the IdP was challenging. This logger will facilitate better understand the flow.

It is important to note that the `idp` logger may expose sensitive information in the logs, such as the `client_secret` or the tokens being issued. Use with care and do not enable it by default. To make it easier to set default log levels and just override some, I've changed a bit the behavior of the logging settings, to allow things like: `all:debug,idp:none`. Previously, if the `all` logger was set, all the other log levels were ignored. Now, if the `all` logger is present, it will be used as the default, then individual loggers will be configured on top.

### DIFF
--- a/e2e/istio/istio_test.go
+++ b/e2e/istio/istio_test.go
@@ -39,7 +39,7 @@ func (i *IstioSuite) TestIstioEnforcement() {
 			// Initialize the test OIDC client that will keep track of the state of the OIDC login process
 			// Initialize it for each test to not reuse the session between them
 			client, err := e2e.NewOIDCTestClient(
-				e2e.WithLoggingOptions(i.T().Log, true),
+				e2e.WithLoggingOptions(i.T().Log),
 				e2e.WithCustomCA(testCAFile),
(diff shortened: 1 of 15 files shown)


sergicastro (commented):
Overall LGTM.

But I have some doubts about `all:debug,idp:none`, weren't we accepting the keyword `default`?  I think it would apply better to this case: `default:debug,idp:none`

nacx:
Not really. [default](https://github.com/istio-ecosystem/authservice/blob/main/internal/logging.go#L33) is an actual logging scope, that is used when no other loggers are configured. It could be misleading, yes... We could consider removing that logger to make this more understandable? But probably in another PR

sergicastro:
Ok, let's keep track of the default keyword in a separate PR.

sergicastro (approved):
Thanks for this.

PR #251, opened 2024-04-17
Cache well-known responses to avoid making too much calls to the IdP

nacx:
While working on https://github.com/istio-ecosystem/authservice/pull/250, I saw that authservice was calling many times the well-known endpoint for a single OIDC session. This is because we call it when initializing the OIDC handlers, and that is done per request to the Envoy ext-authz filter, and a single OIDC session may involve several requests, given the redirects, making several unnecessary calls to the IdP.

This PR caches the seen well-known requests, and returns them from the cache for subsequent requests. The well-known endpoints are usually stable and it is very rare that they change, so it should be fine to just cache them by default.


sergicastro (approved):
Nice

PR #252, opened 2024-04-18
Add minimal examples to make it easier to get started

nacx:
The only configs we had are the `e2e` ones, but it may be hard to get there when you're just trying to get started.
Having a minimal set of configs that are easy to copy/paste will be very helpful.

### DIFF
--- a/examples/minimal/authservice-config.yaml
+++ b/examples/minimal/authservice-config.yaml
@@ -1,0 +1,68 @@
+# Copyright 2024 Tetrate
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


bburky:
I would suggest including a RequestAuthentication and ALLOW AuthorizationPolicy which verify the JWT in the example config. Maybe include a comment to inform users that the CUSTOM AuthorizationPolicy runs first and these run afterwards.

Prior to https://github.com/istio-ecosystem/authservice/pull/187 and https://github.com/istio-ecosystem/authservice/pull/223, this was really the only way to deploy authservice safely: various requests would be allowed through authservice in undesired ways, mostly bypassing authentication. Having Istio perform JWT validation after authservice mitigated this and was the previous recommendation from the authservice developers: https://github.com/istio-ecosystem/authservice/issues/140#issuecomment-1197085674.

I _think_ all of those old issues have been resolved now, and this would now just be extra defense in depth. (If the official recommendation has changed and Istio validation of the JWT is no longer considered required, some updated documentation would be appreciated.)

nacx:
Sounds good. I'll add those examples!

nacx:
Done

sergicastro (approved):
LGTM

PR #257, opened 2024-05-13
Update protoc-gen-go comment to fix `make check`

sergicastro:
When we update the `proto-gen-go` dependency we should be running `make generate` to update the version comment in the generated `.pb.go` files.


nacx (commented):
Can we also pin `protoc-gen-go` to avoid this happening randomly? Let's change this line: https://github.com/istio-ecosystem/authservice/blob/f2ef0e41de8a2d664b35a51bd43748e0ca4d10e8/config/buf.gen.yaml#L23

to `plugin: buf.build/protocolbuffers/go:v1.34.1`

sergicastro:
Good point, version pinned!

nacx (approved):

PR #258, opened 2024-05-14
Fix flaky file watcher test

sergicastro:
The test method `waitForRead` was sometimes signaling due to accumulated reads, even it was inteded to wait for a read after a new content was set in the mock file.

This fixes it by waiting to singal read happened when the read content is the same as the expected one.


sergicastro:
I will join both methods to a single one (set and wait).

nacx (approved):

PR #265, opened 2024-10-09
Add support for PKCE

gdasson:
This PR adds support for PKCE (Proof Key for Code Exchange) to enhance the security of the OAuth 2.0 Authorization Code flow. It includes generating a code verifier and code challenge, and ensures these parameters are used during the authorization and token exchange processes. This update aligns with OAuth 2.0 best practices and improves compatibility with identity providers that require PKCE.

More details about PKCE here: 
* https://oauth.net/2/pkce/
* https://blog.postman.com/what-is-pkce/


internal/oidc/session_test.go
```diff
@@ -133,7 +133,7 @@ func TestSessionGenerator(t *testing.T) {
 		require.NotEqual(t, sg.GenerateNonce(), sg.GenerateNonce())
 	})
 	t.Run("static", func(t *testing.T) {
-		sg := NewStaticGenerator("sessionid", "nonce", "state")
+		sg := NewStaticGenerator("sessionid", "nonce", "state", "codeverifier")
 		require.Equal(t, sg.GenerateSessionID(), sg.GenerateSessionID())
 		require.Equal(t, sg.GenerateState(), sg.GenerateState())
 		require.Equal(t, sg.GenerateNonce(), sg.GenerateNonce())
```
nacx, line 139: [nit] Update this test and the random one to also include a check for the code verifier.

nacx (commented):
Thanks! This looks great. A couple things:

* Can you run `make format`? This should format the code and have CI pass the `check` step.
* One of the `e2e` test suites uses Keycloak. Is there a way to configure it to force that a client uses PKCE? It would be great to have it enforced to make sure we're properly testing that path. Or we could create an additional suite, keycloak-pkce with this setup. WDYT?

internal/oidc/session_test.go
gdasson: @nacx : Thanks for the review. Yes, there is a way to configure keycloak to enforce PKCE use by client. Infact, I also used keycloak for my local testing. In the UI, It's available under Client > Advanced > Proof Key for Code Exchange Code Challenge Method > Set it to S256. (I am not sure if the same setting is available through the Keycloak admin API)

Here is a screenshot from the UI:
image

nacx: The client is created and configured here: https://github.com/istio-ecosystem/authservice/blob/main/e2e/keycloak/setup-keycloak.sh#L50-L58
Let's see if PKCE can be configured in that script.

Actually, there is probably no need to create another suite. The `istio` suite also uses Keycloak, so we can just change the `keycloak` suite and we should have all cases covered.

Testing ti locally should be easy:

```
make docker   # the e2e tests need the update Docker image
make e2e/keycloak

# Or, if you want to see all requests/responses for the e2e tests logged:
E2E_TEST_OPTS="-v -count=1" make e2e/keycloak
```

nacx: Adding this to the keycloak setup script sould do the trick, to enforce PKCE server-side:
```diff
diff --git a/e2e/keycloak/setup-keycloak.sh b/e2e/keycloak/setup-keycloak.sh
index 7fc46af..ee3d340 100755
--- a/e2e/keycloak/setup-keycloak.sh
+++ b/e2e/keycloak/setup-keycloak.sh
@@ -51,6 +51,7 @@ set -ex
     -s clientId="${CLIENT_ID}" \
     -s secret="${CLIENT_SECRET}" \
     -s "redirectUris=[\"${REDIRECT_URL}\"]" \
+    -s "attributes={\"pkce.code.challenge.method\":\"S256\"}" \
     -s consentRequired=false \
     --server "${KEYCLOAK_SERVER}" \
     --realm "${REALM}" \
```

gdasson: Added and ran e2e tests, they run fine.

nacx: Nice! I run the e2e with this setting but without the changes in the PR and they were failing, complaining about the missing query params, so all good!

nacx:
CI complains about project not being go mod tidied.

Since we have code generation, the `check` job makes sure there are no changes after regrating/formatting everything, and fails early if that's not the case.

If you run locally `make check` locally, it should update what's missing and you can commit that to have that check pass.

nacx (approved):
Thanks! Changes LGTM. Let's cleanup the build to have all tests pass.

PR #289, opened 2025-06-22
added post client auth method

ilgatnau:
https://openid.net/specs/openid-connect-core-1_0.html#ClientAuthentication

As per specs, multiple available Client Authentication methods that are used by Clients to authenticate to the Authorization Server when using the Token Endpoint. During Client Registration, the RP (Client) MAY register a Client Authentication method. If no method is registered, the default method is client_secret_basic.

These Client Authentication methods are:

- client_secret_basic
- client_secret_post
- client_secret_jwt
- private_key_jwt

This PR implements implementation for client_secret_post (currently only basic possible) plus placeholder for missing other 2 methods.


config/v1/oidc/config.proto
```diff
@@ -129,6 +129,21 @@ message OIDCConfig {
...
+    CLIENT_AUTHENTICATION_METHOD_UNSPECIFIED = 0;
+    CLIENT_AUTHENTICATION_METHOD_CLIENT_SECRET_POST = 1;
+    CLIENT_AUTHENTICATION_METHOD_CLIENT_SECRET_JWT = 2;
+    CLIENT_AUTHENTICATION_METHOD_PRIVATE_KEY_JWT = 3;
+    CLIENT_AUTHENTICATION_METHOD_BASIC = 4;
+  }
+
+  oneof client_authentication {
+    ClientAuthenticationMethod method = 30;
+  }
```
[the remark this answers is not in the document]
nacx: Also, proto field numbers are a bit random. Can we pick one in sequence of the current ones?

internal/authz/oidc.go
```diff
@@ -331,18 +331,18 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
...
+		return
 	}
 
-	// build headers
-	headers := http.Header{
-		inthttp.HeaderContentType:   []string{inthttp.HeaderContentTypeFormURLEncoded},
-		inthttp.HeaderAuthorization: []string{inthttp.BasicAuthHeader(o.config.GetClientId(), o.config.GetClientSecret())},
+	form, error := buildAuthParams(o.config, codeFromReq, stateFromReq)
+	if error != nil {
+		log.Error("error building form", error)
```
nacx, line 343:
```suggestion
		log.Error("error building auth params", error)
```

internal/authz/oidc.go
```diff
@@ -396,6 +396,81 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
...
 }
 
+// buildAuthHeader builds the authorization header for the client according to the
+// client authentication method specified in the OIDCConfig.
+//
+// The function returns an error if the client authentication method is unspecified
+// or if the implementation for the specified method is not supported.
+func buildAuthHeader(config *oidcv1.OIDCConfig) (http.Header, error) {
+
+	headers := http.Header{}
```
nacx, line 406: There is no need to allocate something here if it is going to be replaced.
```suggestion
	var headers http.Header
```

internal/authz/oidc.go
```diff
@@ -396,6 +396,81 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
 	setDenyResponse(resp, deny, codes.Unauthenticated)
 }
 
+// buildAuthHeader builds the authorization header for the client according to the
+// client authentication method specified in the OIDCConfig.
+//
+// The function returns an error if the client authentication method is unspecified
+// or if the implementation for the specified method is not supported.
+func buildAuthHeader(config *oidcv1.OIDCConfig) (http.Header, error) {
+
```
nacx, line 405:
```suggestion
```

internal/authz/oidc.go
```diff
@@ -396,6 +396,81 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
...
+	headers := http.Header{}
+	switch config.GetMethod() {
+	case oidcv1.OIDCConfig_CLIENT_AUTHENTICATION_METHOD_BASIC:
```
nacx, line 408: Take care. In Go there is no automatic fallthrough! the `default` block will _NOT_ be executed if the client config method is set to basic.

internal/authz/oidc.go
```diff
@@ -396,6 +396,81 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
...
+
+		// Build post auth header
+		headers = http.Header{
+			inthttp.HeaderContentType: []string{inthttp.HeaderContentTypeFormURLEncoded},
+		}
+
+	case oidcv1.OIDCConfig_CLIENT_AUTHENTICATION_METHOD_CLIENT_SECRET_JWT:
+		// Build jwt auth header
+		// TODO: implement jwt auth header
+		return nil, errors.New("not implemented")
```
nacx, line 425: Add more details in the error about _what_ is not implemented so users can understand what is the problem when receiving the response back.

internal/authz/oidc.go
```diff
@@ -396,6 +396,81 @@ func (o *oidcHandler) retrieveTokens(ctx context.Context, log telemetry.Logger,
...
+	}
+
+	return headers, nil
+}
+
+func buildAuthParams(config *oidcv1.OIDCConfig, codeFromReq string, codeVerifierFromReq string) (url.Values, error) {
+	params := url.Values{}
+	switch config.GetMethod() {
+	case oidcv1.OIDCConfig_CLIENT_AUTHENTICATION_METHOD_BASIC:
+	default:
```
nacx, line 440: Same two comments as in the previous method.

ilgatnau:
> Thanks!
> 
> I've left some comments on the PR. Thanks for raising!
> 
> In order to have the PR merged and in s state that can be safely maintained a couple things need to be added:
> 
> * Unit tests validating that the POST method works as expected. The current PR only configures changes tests for the BASIC. Make sure there are unit tests for the case where no method is set, and when basic and post are set. The three cases need to be explicitly covered.
> * An end-to-end test that also exercises the post method. We use Keycloak in the e2e tests. Some suite should be updated to use the post mehtod and verify it is working as expected.
> 
> Having these two points properly implemented and covered will help us maintain the feature without breaking it accidentally in the future.
> 
> Thanks!

Thank you for your comments @nacx, while I am trying to implement new tests for new client authentication methods I have some doubts how you guys would like to have this done as TestOIDCProcess looks like an uber test difficult to breakdown in more atomic tests.

I have for now duplicated this test as a reusable function I can call with different client authentication methods, would that be acceptable?

See commit https://github.com/ilgatnau/authservice/commit/998f12b45eaf71b414a82710a9f19cc287a14955

internal/authz/oidc_test.go
```diff
@@ -940,6 +1012,793 @@ func TestOIDCProcess(t *testing.T) {
...
+	t.Run("callback request ", func(t *testing.T) {
+		resp := &envoy.CheckResponse{}
+		require.NoError(t, h.Process(ctx, callbackRequest, resp))
+		require.Equal(t, int32(codes.Unauthenticated), resp.GetStatus().GetCode())
+		requireStandardResponseHeaders(t, resp)
+		requireStoredTokens(t, store, sessionID, false)
+	})
+}
+
+func testOIDCProcessReusable(t *testing.T, oidcConfig *oidcv1.OIDCConfig) {
```
nacx, line 1067: What is the main difference between this method and the existing `TestOIDCProcess` one? Is it that this one just gets the OIDC config as an argument?
Can we refactor the tests so that we reuse the code and add these cases to a single one, instead of duplicating the test logic?

PR #293, opened 2025-07-29
Generalize secret controller

nacx:
Makes the k8s secret controller more generic by:

* Not restricting it to the current namespace. Right now, it's only watching the current namespace and ignores the `namespace` that users set in the configuration. This change properly honors user config and only watches for changes in the configured namespaces.
* Not coupling the reconciler to the `client-secret` key in a secret. Instead, the update functions will get the entire secret data, so the same controller can be used to update the config for other secrets (like future TLS secrets for the redis store, etc).


internal/k8s/secret_controller.go
```diff
@@ -73,42 +73,27 @@ func (s *SecretController) Name() string { return "Secret controller" }
...
-			if isOIDCConf &&
-				(oidcCfg.Oidc.GetClientSecretRef() != nil || oidcCfg.Oidc.GetTokenExchange().GetClientCredentials().GetClientSecretRef() != nil) {
-				needWatchSecrets = true
-				break
-			}
+	if s.defaultNamespace == "" {
+		var data []byte
+		data, err := os.ReadFile(NamespacePath)
+		if err != nil {
+			s.log.Error("error reading namespace file", err, "file", NamespacePath)
```
sergicastro, line 80:
```suggestion
			s.log.Error("error reading namespace file, watching \"default\" namespace", err, "file", NamespacePath)
```

PR #294, opened 2025-08-01
Add options to configure TLS and mTLS connections to Redis

nacx:
This change allows configuring TLS properties for the redis connection as well as the password. The values can be set in plain text in the configuration or as references to files.

Note that even though the content of the Redis certificates or password may live in a secret, I think it is more appropriate to reference them as a file, not the secret, because:

* If the secret is mounted in the Authservice, the contents are ready since the Authservice start, removing any eventual consistency until the k8s reconciler has picked up the secret and propagated the updated value.
* It does not require the AuthService to have additional permissions to read secrets in the cluster.

For the Authservice use cases, mounting the secrets is what makes more sense, and I'll propose deprecating the secret reference APIs in favor of local files.


sergicastro (commented):
Overall LGTM, just small nits

sergicastro (approved):

PR #302, opened 2025-09-04
chore: separate the tool dependencies from the main go.mod

nacx:
This PR separates the tools into a different module: `tools/go.mod`. This way, the tool dependencies will not coexist and affect the versions of the dependencies in the main module.


sergicastro (approved):
🙌🏼

PR #342, opened 2026-07-23
feat: add sentinel support

### LINKED ISSUE #325: Support Redis Sentinel for session store (HA-aware client)
## Summary

authservice's Redis session store constructs its client via go-redis's standalone `NewClient` path, which is not Sentinel-aware. In deployments backed by a Sentinel-managed HA Redis (or Valkey) cluster, this means authservice continues talking to the former primary after a Sentinel-driven failover, surfacing a sustained stream of `READONLY You can't write against a read only replica` errors to users until the authservice pods are manually restarted.

Raising this to confirm the problem framing and see whether maintainers would welcome a contribution in this direction.

## Current behavior

`NewRedisClient` at [`internal/oidc/redis.go:71,116`](https://github.com/istio-ecosystem/authservice/blob/main/internal/oidc/redis.go) parses the `server_uri` and hands it to `redis.NewClient`:

```go
opts, err := redis.ParseURL(config.GetServerUri())
// ...
return redis.NewClient(opts), nil
```

`redis.ParseURL` only accepts `redis://`, `rediss://`, and `unix://` ([go-redis v9.7.3 `options.go:298-305`](https://github.com/redis/go-redis/blob/v9.7.3/options.go#L298-L305)). The standalone client has no mechanism to discover a new primary on its own.

Consequence on Sentinel-driven failover:

1. Former primary is demoted and starts rejecting writes with `READONLY You can't write against a read only replica`.
2. The standalone client surfaces the error to authservice with no Sentinel-aware retry or re-resolution.
3. Pool-level teardown eventually forces a reconnect, but DNS re-resolution on the same Service name lands back on the same endpoints.
4. Writes remain broken until authservice pods are restarted.

Related issue [#240](https://github.com/istio-ecosystem/authservice/issues/240) (still open) describes the same class of breakage attributed to replica sync lag; PR [#241](https://github.com/istio-ecosystem/authservice/pull/241) attempted a fix and was closed. I suspect the root cause of both is this standalone-client construction.

## Proposed direction

Use go-redis's built-in Sentinel client. [`NewFailoverClient`](https://github.com/redis/go-redis/blob/v9.7.3/sentinel.go#L225) returns the same `*redis.Client` type as `NewClient`, so the rest of the code (via `redis.Cmdable` at `internal/oidc/redis.go:62`) is unchanged. The failover client [subscribes to Sentinel's `+switch-master` channel](https://github.com/redis/go-redis/blob/v9.7.3/sentinel.go#L740) and [updates its internal master address on each event](https://github.com/redis/go-redis/blob/v9.7.3/sentinel.go#L779-L786) — no additional client-side logic needed.

The API is already available in the `go-redis/v9 v9.7.3` authservice is pinned on, so this wouldn't require a dependency bump. [oauth2-proxy uses this exact pattern](https://github.com/oauth2-proxy/oauth2-proxy/blob/master/pkg/sessions/redis/redis_store.go) (its `buildSentinelClient`) and has shipped it in production for years.

Config-wise, an optional Sentinel sub-message on `RedisConfig` would be additive and preserve the existing `server_uri` path.

ntwkninja:
### Summary
When `server_uri` uses the `redis+sentinel://` (or `rediss+sentinel://`) scheme, the Redis session store now connects through go-redis's Sentinel-aware `NewFailoverClient`, which tracks the current primary across Sentinel-driven failovers and eliminates the sustained `READONLY You can't write against a read only replica` errors that previously persisted until authservice pods were restarted. 

The URI is parsed by go-redis's own `redis.ParseFailoverURL` (authservice defines only the `+sentinel` scheme marker and strips it before delegating) so URL semantics are exactly go-redis's documented failover-URL form and stay in sync with the library: `redis+sentinel://[sentinel-user:sentinel-pass@]host:26379[/db]?master_name=mymaster[&addr=host2:26379]`. URL userinfo authenticates to the sentinels (`SentinelUsername`/`SentinelPassword`); master credentials come from `username`/`password` query params or the existing `username`/`password`/`password_file` config fields, which overlay the URL exactly as they do for the standalone client. Config load validates the URI and requires `master_name`. 

Opting in via a URI scheme on the existing `server_uri` keeps the config API and the standalone `redis://`,`rediss://`,`unix://` paths unchanged, and adds no dependencies. `NewFailoverClient` and `ParseFailoverURL` both ship in the go-redis v9.20.0 (already pinned here). Regression tests pin all legacy schemes to the standalone client. 

### Reviewer Note
The one deliberate deviation from upstream parsing is clearing `TLSConfig.ServerName` after parse; go-redis pins it to the first sentinel host but reuses that TLS config when dialing the master, so leaving it empty restores correct per-connection hostname verification.

closes #325


internal/oidc/redis.go
```diff
@@ -68,6 +68,50 @@ type redisStore struct {
...
+				failoverOpts.TLSConfig.RootCAs = caCertPool
+			}
+			if certPEM := tlscfg.GetClientCertPem(); certPEM != "" {
+				cert, err := tls.X509KeyPair([]byte(certPEM), []byte(tlscfg.GetClientKeyPem()))
+				if err != nil {
+					return nil, fmt.Errorf("loading client certificate: %w", err)
+				}
+				failoverOpts.TLSConfig.Certificates = []tls.Certificate{cert}
+			}
+		}
```
nacx, line 102: Can you move this to a small helper and reuse it as well in the code below, to avoid repetition?

nacx (commented):
Apologies for the late review. I've been PTO.

I like the approach! The only caveat is that for authenticated sentinels, the only option (IIUC) is to encode the credentials in the URL, and having the password there is not the best.

WDYT about adding a `sentinelUsername` and `sentinelPassword` to the redis config, so users can configure that the same way the configure the current credentials? (URL, inline, via secret).

e2e/sentinel/sentinel.conf
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+port 26379
+
+# Single sentinel with quorum 1, tuned for fast failover in tests
+sentinel monitor mymaster 127.0.0.1 6380 1
+sentinel down-after-milliseconds mymaster 1000
+sentinel failover-timeout mymaster 3000
```
nacx, line 22: Can we have a configuration that requires sentinel auth, to exercise the credentials processing path?

nacx:
Ignore the scan job failure. It will fix after rebasing to latest main

ntwkninja:
> Apologies for the late review. I've been PTO.
> 
> I like the approach! The only caveat is that for authenticated sentinels, the only option (IIUC) is to encode the credentials in the URL, and having the password there is not the best.
> 
> WDYT about adding a `sentinelUsername` and `sentinelPassword` to the redis config, so users can configure that the same way the configure the current credentials? (URL, inline, via secret).

Thanks @nacx, hope you enjoyed your time off! Totally agree with your assessment. Pushed up the changes to address.

internal/oidc/redis_test.go
```diff
@@ -213,6 +213,51 @@ func TestRedisMTLS(t *testing.T) {
...
+		},
+	})
+	require.NoError(t, err)
+
+	rc, ok := client.(*redis.Client)
+	require.True(t, ok)
+	// go-redis sets Addr to "FailoverClient" on failover clients
+	require.Equal(t, "FailoverClient", rc.Options().Addr)
+	require.Equal(t, "user", rc.Options().Username)
+	require.Equal(t, "pass", rc.Options().Password)
```
nacx, line 240: Verify the sentinel credentials as well?

nacx (commented):
Overall LGTM. Thanks!
Mind fixing the DCO so we can merge?

internal/oidc/redis_test.go
```diff
@@ -213,6 +213,51 @@ func TestRedisMTLS(t *testing.T) {
...
+}
```
nacx, line 244: Can you add a test that verifies that, with Sentinel and TLS, the client is configured with an empty ServerName in the TLS Config? This looks important according to the comments, but there is no unit test guarding it.

internal/oidc/redis_test.go
ntwkninja: go-redis does not expose Sentinel credentials through Client.Options(). I added an e2e assertion that unauthenticated Sentinel access fails with NOAUTH; the authenticated store and failover tests verify the configured credentials functionally.

internal/oidc/redis_test.go
ntwkninja: Added assertions that standalone TLS sets ServerName to the Redis host, while Sentinel TLS leaves it empty for per-connection derivation and keeps certificate verification enabled.

nacx (approved):
LGTM. Thank you!

